PR #1, opened 2020-03-30
Adds another friendly message to hello_world.py

jgoodknight:
Adds another friendly message to hello_world.py

### DIFF
--- a/BUILD
+++ b/BUILD
@@ -1,0 +1,6 @@
+# Google's Cross-Publisher Reach Experiments
+package(default_visibility = ["//visibility:public"])
+
+licenses(["notice"])
+
+exports_files(["LICENSE"])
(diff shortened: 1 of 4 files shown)


BUILD
```diff
@@ -0,0 +1,6 @@
+# Google's Cross-Publisher Reach Experiments
```
kungfucraig: I don't think we want to mention Google in this.

rmirisola: Let's also wait for formal confirmation Bazel is ok to use.

jgoodknight: I will definitely remove this!  Our internal tooling does not sync README files by default so oddly enough our internal readme does not mention google.  I will dig into this and figure out the best way forward, and push a new commit to this branch shortly.  Thank you!

jgoodknight: just realized this is a BUILD file actually, will remove it from this commit and do a more earnest BUILD migration later if we do use bazel.  stand by

PR #6, opened 2020-06-30
Project import generated by Copybara.

jgoodknight:
Project import generated by Copybara.


src/common/noisers.py
```diff
@@ -0,0 +1,114 @@
...
+    return np.round(self._func_with_noise(x))
+
+
+def main(argv):
+  if len(argv) > 1:
+    raise app.UsageError('Too many command-line arguments.')
+
+
+if __name__ == '__main__':
+  app.run(main)
```
pfhgetty, line 114: Does this really run? Where is app imported?

src/estimators/estimator_noisers.py
```diff
@@ -0,0 +1,74 @@
...
+      return self._noiser(cardinality_estimate)
+
+
+def main(argv):
+  if len(argv) > 1:
+    raise app.UsageError('Too many command-line arguments.')
+
+
+if __name__ == '__main__':
+  app.run(main)
```
pfhgetty, line 74: Same for this one

src/estimators/tests/bloom_filters_test.py
```diff
@@ -120,6 +120,12 @@ def test_multi_insertion_same_element(self):
     b1.add(1)
     self.assertEqual(UnionEstimator.estimate_cardinality(b1), 1)
 
+  def test_raise_error_with_full_bloom_filter(self):
+    b1 = BloomFilter(length=2, random_seed=3)
+    b1.add_ids(range(10))
+    with self.assertRaises(ValueError):
+      _ = UnionEstimator.estimate_cardinality(b1)
```
pfhgetty, line 127: This is confusing syntax. Why are we assigning a throwaway to the return of this function? Shouldn't we just call the function instead without assigning anything?

src/common/noisers.py
jgoodknight: yeesh, good catch.  Normally we would do 'from absl import app' but this doesn't need to be a executable anyway

src/estimators/estimator_noisers.py
jgoodknight: done

src/estimators/tests/bloom_filters_test.py
jgoodknight: another good call, will remove

jgoodknight:
I think we messed something up internally, my apologies.  Will have to regenerate this PR

PR #10, opened 2020-07-07
Google's last upload using copybara

jgoodknight:
Google's last upload using copybara


setup.py
```diff
@@ -24,6 +24,6 @@
         'wfa_cardinality_estimation_evaluation_framework',
         'wfa_cardinality_estimation_evaluation_framework.common',
         'wfa_cardinality_estimation_evaluation_framework.estimators',
+        'wfa_cardinality_estimation_evaluation_framework.evaluations',
         'wfa_cardinality_estimation_evaluation_framework.simulations',
```
my2457, line 28: Probably also need to add 'wfa_cardinality_estimation_evaluation_framework.evaluations.data'?

src/simulations/simulator.py
```diff
@@ -158,24 +150,26 @@ def run_one(self):
...
       actual_ids.append(campaign_ids)
-      sketch = self.estimator_config.sketch_factory(sketch_random_seed)
+      sketch = self.sketch_estimator_config.sketch_factory(sketch_random_seed)
       sketch.add_ids(campaign_ids)
       sketches.append(sketch)
 
-    # Optionally noise the sketches if it exists in the estimator_config.
-    if hasattr(self.estimator_config, 'sketch_noiser') and self.estimator_config.sketch_noiser:
+    # Optionally noise the sketches.
+    if hasattr(self.estimator_config,
```
my2457, line 158: self.sketch_estimator_config?

src/evaluations/tests/evaluator_test.py
```diff
@@ -178,7 +177,7 @@ def test_evaluate_all_save_results(self):
               f'{sketch_estimator_config.name}/{scenario_config.name}')
 
   def test_same_random_state_of_same_scenario(self):
-    test_evaluator = self.get_test_evaluator(self.create_tempdir().full_path)
+    test_evaluator = self.get_test_evaluator(self.create_tempdir())
```
pfhgetty, line 180: Why was this changed? This change breaks causes TypeErrors on my machine.

src/evaluations/tests/analyzer_test.py
```diff
@@ -138,6 +139,32 @@ def test_get_num_estimable_sets_df_results_correct(self):
     except AssertionError:
       self.fail('The number of estimable sets is not correct.')
 
+  def test_get_relative_error_stats_of_num_of_estimable_sets(self):
```
pfhgetty, line 142: Line 58 and 64 are broken by changes set_generator.py changes. I think get_generator_factory_with_num_and_size is intended instead of get_generator_factory but I'm not 100% sure.

setup.py
jgoodknight: I don't think so.  We decided to keep the data module not as a package so you won't have to re-install the package every time you change code in data, as we expect data to change frequently

src/simulations/simulator.py
jgoodknight: sorry I'm not sure what you're suggesting: can you clarify?

src/evaluations/tests/evaluator_test.py
jgoodknight: I must have missed this when copying things over manually: my apologies.

setup.py
jgoodknight: my apologies, Preston pointed out a problem which I tracked to a misconfiguration of our github importer.  Hopefully the upcoming upload makes more sense

src/simulations/simulator.py
my2457: Did you actually mean self.sketch_estimator_config rather than  self.estimator_config here? I saw self.estimator_config was never set in any method and this access would probably throw AttributeError?

src/evaluations/tests/analyzer_test.py
jgoodknight: i think this has been addressed, let me know

src/simulations/simulator.py
jgoodknight: there was a merge error I should have caught, the new commit should address this.  Thanks!

evaluations/run_evaluation.py
```diff
@@ -28,16 +28,24 @@
 
 from wfa_cardinality_estimation_evaluation_framework.evaluations import analyzer
 from wfa_cardinality_estimation_evaluation_framework.evaluations import evaluator
-from data import evaluation_configs
+from wfa_cardinality_estimation_evaluation_framework.evaluations import report_generator
+from wfa_cardinality_estimation_evaluation_framework.evaluations.data import evaluation_configs
```
pfhgetty, line 32: This should be:
from data import evaluation_configs
because data is not part of the evaluations package anymore.

src/evaluations/report_generator.py
```diff
@@ -0,0 +1,286 @@
...
+import shutil
+
+import pandas as pd
+
+from wfa_cardinality_estimation_evaluation_framework.evaluations import analyzer
+from wfa_cardinality_estimation_evaluation_framework.evaluations import evaluator
+from wfa_cardinality_estimation_evaluation_framework.evaluations.analyzer import KEY_DESCRIPTION_TO_FILE_DIR
+from wfa_cardinality_estimation_evaluation_framework.evaluations.analyzer import KEY_NUM_ESTIMABLE_SETS_STATS_DF
+from wfa_cardinality_estimation_evaluation_framework.evaluations.analyzer import KEY_RUNNING_TIME_DF
+from wfa_cardinality_estimation_evaluation_framework.evaluations.data import evaluation_configs
```
pfhgetty, line 26: data is not part of the package. This file needs to be moved out into the top-level evaluations/ directory.
(or the whole evaluations folder should be in src/. Either one will do.)

jgoodknight: done, and confirmed run_evaluation works from git clone (although I didn't set the required flags so it didn't fully run, just was able to import things)

src/evaluations/run_evaluation.py
```diff
@@ -58,69 +66,79 @@
...
+    'proportion of runs. By default, use 0.95.')
+flags.DEFINE_integer('boxplot_xlabel_rotate', 0,
+                     'The degrees of rotation of the x labels in the boxplot.')
 flags.DEFINE_integer('boxplot_size_width_inch', 12,
                      'The widths of the boxplot in inches.')
 flags.DEFINE_integer('boxplot_size_height_inch', 6,
                      'The widths of the boxplot in inches.')
 
+required_flags = ('evaluation_config', 'sketch_estimator_configs',
+                  'evaluation_run_name', 'num_runs')
```
my2457, line 85: Shall we also add `evaluation_out_dir` to required?

src/evaluations/analyzer.py
```diff
@@ -107,16 +120,19 @@ def __init__(self, out_dir, evaluation_directory, evaluation_run_name,
...
-    self.proportion_of_runs = proportion_of_runs
-    self.plot_params = plot_params or PLOT_PARAMS
+    self.estimable_criteria_list = estimable_criteria_list
+    if plot_params is None:
+      self.plot_params = PLOT_PARAMS
+    else:
+      self.plot_params = plot_params
 
     # Get all the raw results.
     self.evaluation_file_dirs = evaluator.load_directory_tree(
```
my2457, line 138: Seems `evaluation_directory` is passed here, but  `os.path.join(out_dir, run_name)` in `load_directory_tree` would throw exception if it's null

jgoodknight: sure good idea

src/evaluations/run_evaluation.py
jgoodknight: sure good idea!

src/evaluations/analyzer.py
jgoodknight: true but since there's not a default value in the constructor I don't think it's a worry that evaluation_directory would ever be None

PR #13, opened 2020-07-15
Updates EstimatorBase to return a histogram of estimates.

matthewclegg:


requirements.txt
```diff
@@ -4,3 +4,5 @@ matplotlib==3.0.3
 numpy==1.16.4
 scipy==1.2.1
 seaborn==0.9.0
+lxml==4.5.2
```
huangxichen1, line 7: If my understanding is correct, it is needed by the report_generator_test.py right?

matthewclegg: Yes, I believe that's correct.

src/estimators/exact_set.py
```diff
@@ -83,7 +83,8 @@ def __init__(self):
   def __call__(self, sketch_list):
     """Return len(sketch)."""
```
kungfucraig, line 84: Update doc string

matthewclegg: Done.

src/estimators/exact_set.py
```diff
@@ -83,7 +83,8 @@ def __init__(self):
...
     e = LosslessEstimator()
-    return e(sketch_list) - 1
+    histogram = e(sketch_list)
+    return [histogram[0]-1] + [histogram[1:]]
```
my2457, line 87: Why do we want to append `[histogram[1:]]`? If `e` returns true cardinality, it should be 
`[x-1 for x in histogram]` ?

matthewclegg: Thank you for the comment.  First, it was an undetected error to enclose histogram[1:] in a list.   Also, I seem to have overlooked the corner case when histogram[0] == 0.  (Ack!! Two bugs in one line of code! I hate it when that happens!)

As to your question, I guess it depends on how you interpret the phrase "always short by one".  I interpreted this to mean that the LessOneEstimator should provide a cardinality estimate that is one smaller than the cardinality estimate provide by LosslessEstimator.  In other words, if the total cardinality returned by LosslessEstimator is n, then the total cardinality returned by LessOneEstimator should be n-1.  If we make the change you suggested, then the total cardinality returned by LessOneEstimator will be n - f * (f + 1) / 2, where f is the maximum frequency. 

Let me know your thoughts.

my2457: Not sure I understand this correctly, if we have `[x-1 for x in histogram]` (suppose x>0 `histogram[-1]`) then cardinality estimate is still of by 1 since `ret[0]` is the true cardinality with `f >=1`? But the problem this way is that we're always missing the one with the highest frequency, so probably better to have a random "pivot" where all indexes below is off by 1 but indexes above are the same?

matthewclegg: Actually true cardinality is sum([(i+1) * h[i] for i in range(len(h))]).  In other words, h[0] is the number of id's occurring exactly once.  So, let's say the list of id's is [1, 2, 2, 3, 4, 4, 5, 6, 6, 6].  Then, the corresponding value of histogram is 
[3, 2, 1].  As I've coded LessOneEstimator, it would then return [2, 2, 1].

my2457: From the spec here, `If h[] is the list of values returned by EstimatorBase, then h[i] is defined as the number of id's that occurred with a frequency of i+1 or higher`

And I just saw the comment in `EstimatorBase` that seems to be conflicting with what mentioned in docs. So which route are we going to take?

https://github.com/world-federation-of-advertisers/cardinality_estimation_evaluation_framework/blob/freq-eval-design/doc/frequency_evaluation_framework.md#sketchestimator

src/estimators/base.py
```diff
@@ -51,10 +51,18 @@ def f(random_state):
...
-  """An estimator takes a sketch and produces a cardinality estimate.
+  """An estimator takes a sketch and produces a frequency histogram.
+
+  If the array of values produced by an estimator is h[], then h[i] represents
+  the cardinality estimate of items having a frequency of i+1.  h[-1] is the
+  cardinality estimate of items having a frequency of len(h) or greater.
+
+  Note that there is no requirement on the number of frequencies that
+  an estimator should produce.  Thus, an estimator that provides estimates
+  only of cardinality may output an array of length 1.
```
my2457, line 62: I think this is the part that's contradicting the doc, but if this is the case, we can't simply use `EstimatorBase[0]` as cardinality estimate?

src/estimators/tests/bloom_filters_test.py
```diff
@@ -157,7 +157,7 @@ def test_cardinality_estimation(self):
     b2 = BloomFilter(length=10, random_seed=3)
     b2.sketch[1] = 1
 
-    cardinality = UnionEstimator()([b1, b2])
+    cardinality = UnionEstimator()([b1, b2])[0]
```
my2457, line 160: Like all the tests here, even though they may return the same result, I would say `sum(UnionEstimator()([b1, b2]))` would make more sense?

src/estimators/exact_set.py
matthewclegg: Thank you very much for pointing this out.  I have updated the definition of EstimatorBase to match the doc.

src/estimators/base.py
matthewclegg: Yes, you are right.

src/estimators/tests/bloom_filters_test.py
matthewclegg: Is this comment now moot?

PR #15, opened 2020-07-16
Speed Improvement By Replacing np.random.choice

pfhgetty:
np.random.choice with replace=False allocates large amounts of memory linear with number of items to choose from. This PR replaces np.random.choice with Robert Floyd's No-Replacement Sampling algorithm which is linear with how many items we are choosing (which is generally a much smaller number than the items we can choose from for our use case).

-- This method is nearly 1000x faster when the population is greater than 10^9
-- The runtime of the simulation for vector_of_counts-4096-ln3-sequential on the smoke_test scenarios is about 3 seconds with this fix on a Google Pixelbook. Without this fix, the runtime is over 3 minutes.


src/simulations/set_generator.py
```diff
@@ -25,6 +25,7 @@
 from absl import logging
 import numpy as np
 from wfa_cardinality_estimation_evaluation_framework.common.analysis import relative_error
+import collections
```
matthewclegg, line 28: Are we using Google's style guide?  If so, standard library imports should come before others.

src/simulations/set_generator.py
```diff
@@ -51,6 +52,48 @@ def __iter__(self):
     for _ in range(self.num_sets):
       yield self.set_size
 
+def _choice_fast(n, m, random_state):
+    """O(m) space-optimal algorithm for generating m random indices for list
```
matthewclegg, line 56: Best practice is to open with a single line comment, not exceeding column 80, describing the purpose of the function, terminated with a period.

src/simulations/set_generator.py
```diff
@@ -51,6 +52,48 @@ def __iter__(self):
     for _ in range(self.num_sets):
       yield self.set_size
 
+def _choice_fast(n, m, random_state):
+    """O(m) space-optimal algorithm for generating m random indices for list 
+    of size n without replacement
+    
+    Args:
```
matthewclegg, line 61: See the style guide on comments for functions and methods ...

src/simulations/set_generator.py
```diff
@@ -51,6 +52,48 @@ def __iter__(self):
...
+    ### Robert Floyd's No-Replacement Sampling Algorithm ###
+    # Create an empty set to place numbers in
+    s = set()
+    for j in range(size-m, size):
+        t = random_state.randint(0, j + 1)
+        if t not in s:
+            s.add(t)
+        else:
+            s.add(j)
+    assert len(s) == m
```
matthewclegg, line 94: Is this guaranteed by the code?

src/simulations/set_generator.py
```diff
@@ -51,6 +52,48 @@ def __iter__(self):
     for _ in range(self.num_sets):
       yield self.set_size
 
+def _choice_fast(n, m, random_state):
```
matthewclegg: Would it make sense to put this into its own file?  I think that we are going to have a number of different set generation methods, which will probably span multiple files.

src/simulations/set_generator.py
pfhgetty: Yes

src/simulations/set_generator.py
pfhgetty: It could definitely make sense to put this into its own file. My original logic was that this was a niche function that was only going to be used in this file. Currently, I believe that set_generator.py is the only file that contains (multiple) set generation methods so for now I believe the scope of this function should be specific to set_generator.py. If it is guaranteed that we will start separating out the set generation methods into multiple files, let me know! I will definitely consider putting it in a different file if that is the case.

src/simulations/set_generator.py
pfhgetty: Resolved (I think!) in c5bdec7

src/simulations/set_generator.py
pfhgetty: Resolved (I think!) in c5bdec7

src/simulations/set_generator.py
pfhgetty: Resolved in c5bdec7

src/simulations/set_generator.py
```diff
@@ -51,6 +53,57 @@ def __iter__(self):
     for _ in range(self.num_sets):
       yield self.set_size
 
+def _choice_fast(n, m, random_state=np.random.RandomState()):
+    """Chooses m numbers or objects from population n.
+
+    O(m) space-optimal algorithm for generating m random indices for list
```
jgoodknight, line 59: Can you put a reference of some kind for the algorithm into the comment?

src/simulations/set_generator.py
```diff
@@ -51,6 +53,57 @@ def __iter__(self):
...
+    assert isinstance(n, collections.abc.Iterable) or isinstance(n, int)
+    # Get the maximum number as size
+    if isinstance(n, int):
+      size = n
+    else:
+      size = len(n)
+    # We should always be choosing fewer than or up to size
+    assert m <= size
+
+    ### Robert Floyd's No-Replacement Sampling Algorithm ###
```
jgoodknight, line 82: or you could put the reference here.

src/simulations/set_generator.py
```diff
@@ -51,6 +53,57 @@ def __iter__(self):
     for _ in range(self.num_sets):
       yield self.set_size
 
+def _choice_fast(n, m, random_state=np.random.RandomState()):
```
jgoodknight, line 56: I think this would do well in the common directory instead of in the set_generator.  I can't immediately think of anything else using it, but you never know what the future might hold.  At the least I don't think you need to worry about it being private: what made you decide to make it private?

src/simulations/tests/set_generator_test.py
```diff
@@ -414,6 +414,34 @@ def test_sequentially_correlated_all_previous_generator_raise_error(self):
           order='random', correlated_sets='not_implemented', universe_size=30,
           shared_prop=0.2, set_sizes=[10] * 3, random_state=rs)
 
+  def test_choice_fast_same_random_state_same_output(self):
```
jgoodknight, line 417: nice test!

src/simulations/set_generator.py
pfhgetty: Oh, I didn't even realize there was a common directory! That probably would be a better place for it

src/simulations/set_generator.py
pfhgetty: @jgoodknight mentioned that there was a common folder as part of our package that I missed! This method definitely belongs there instead. Fixed in 38fadaa

src/simulations/set_generator.py
pfhgetty: 38fadaa

src/simulations/set_generator.py
pfhgetty: Resolved in 38fadaa

src/simulations/set_generator.py
pfhgetty: Resolved in 38fadaa

PR #17, opened 2020-07-17
Parallel Computation Model

pfhgetty:
This PR introduces parallel computation for the Evaluator object in evaluator.py. The parallelization is handled by pathos, a Python multiprocessing package that uses 'dill' for serialization instead of Python's default 'pickle'.
- Testing 'vector_of_counts-4096-ln3-sequential' on 'smoke_test' on a Pixelbook:
  - Without the (#15) fix:
    - Running serially on a single core takes ~170 seconds
    - Running parallel on two cores (hyperthreaded to 4 virtual cores) takes ~130 seconds
  - With the (#15) fix:
    - Running serially on a single core takes ~2.5 seconds
    - Running parallel on two cores (hyperthreaded to 4 virtual cores) takes ~1.5 seconds
- dill is required due to local functions that we want to pass into our simulations. pickle cannot serialize this type of data.
- tqdm is introduced for dynamic loading bars and time estimates during parallelization.
- tqdm calculates wall time estimates and outputs actual wall time spent. **The 'evaluation_run_time' file inside the evaluation output folder for each estimator should produce approximately the same results whether run parallel or serially. This is by design to allow easy comparison while debugging estimators.** The evaluation_run_time when run parallel will sum up the time spent on all cores to measure CPU time, not wall time.
- Added non-required parallel_cores argument to run_evaluation.py to specify how many cores to use while running evaluations. If not specified, will use all available CPU cores.


src/evaluations/evaluator.py
```diff
@@ -198,7 +203,9 @@ def __init__(self, evaluation_config, sketch_estimator_config_list, run_name,
       assert isinstance(sketch_estimator_config, SketchEstimatorConfig), (
           'sketch_estimator_config_list does not contain all '
           'SketchEstimatorConfig objects')
-
+    assert isinstance(parallel_cores, int), (
+        'parallel_cores must be an int')
+    self.parallel_cores = parallel_cores if parallel_cores >= 1 else cpu_count()
```
matthewclegg, line 208: What happens if the user requests more than the maximum available number of cores?

src/evaluations/evaluator.py
```diff
@@ -224,14 +231,17 @@ def __init__(self, evaluation_config, sketch_estimator_config_list, run_name,
     self.scenario_random_states = scenario_random_states
 
   def __call__(self):
-    self.evaluate_all()
+    if self.parallel_cores > 1:
```
matthewclegg, line 234: Why keep evaluate_all()?

src/evaluations/evaluator.py
pfhgetty: Ah this is a good point. If the user requests more than the maximum cores, then that number of processes are spawned and there will be at least 1 core with more than 1 process. It might be best to cap it to the maximum amount of processors because no performance gains will be had from spawning more than 1 process per core.

src/evaluations/evaluator.py
pfhgetty: This is a good point. I hadn't even thought about it, but I suppose the evaluate_all_parallel method works nearly exactly the same way that evaluate_all works. I will change this to only use evaluate_all_parallel.

pfhgetty: Resolved in 27cfae7

src/evaluations/evaluator.py
pfhgetty: Resolved in 27cfae7

src/evaluations/evaluator.py
```diff
@@ -205,7 +205,12 @@ def __init__(self, evaluation_config, sketch_estimator_config_list, run_name,
           'SketchEstimatorConfig objects')
     assert isinstance(parallel_cores, int), (
         'parallel_cores must be an int')
-    self.parallel_cores = parallel_cores if parallel_cores >= 1 else cpu_count()
+    if parallel_cores > cpu_count():
+      self.parallel_cores = cpu_count()
+    elif parallel_cores >= 1:
+      self.parallel_cores = parallel_cores
+    else:
+      self.parallel_cores = cpu_count()
```
my2457, line 213: probably it's clearer this way?
```suggestion
    self.parallel_cores = min(1, parallel_cores, cpu_count())
```

src/evaluations/evaluator.py
```diff
@@ -198,7 +203,15 @@ def __init__(self, evaluation_config, sketch_estimator_config_list, run_name,
       assert isinstance(sketch_estimator_config, SketchEstimatorConfig), (
           'sketch_estimator_config_list does not contain all '
           'SketchEstimatorConfig objects')
-
+    assert isinstance(parallel_cores, int), (
+        'parallel_cores must be an int')
+    if parallel_cores < 1:
+      self.parallel_cores = cpu_count()
```
kungfucraig, line 209: Assuming this is actually "number of threads to use" then cpu_count() * c  where c > 1 can be nice to help account of I/O boundedness of some tasks. Typically "c" would be an input param. We don't necessarily need to add this though.

It's unclear to me after a brief look at pathos docs whether this is actually #cores or #threads though. Can you clarify?

src/evaluations/evaluator.py
```diff
@@ -14,14 +14,16 @@
...
+import itertools
 import os
 import pickle
 import shutil
 import time
 
 from absl import logging
-
 import numpy as np
+from pathos.multiprocessing import ProcessPool, cpu_count
```
kungfucraig, line 25: Why pathos and not the native multiprocessing library?

src/evaluations/evaluator.py
```diff
@@ -224,38 +237,94 @@ def __init__(self, evaluation_config, sketch_estimator_config_list, run_name,
...
+    # Spawnable process to evaluate a sketch estimator on a certain scenario.
+    def run_scenario(args):
+      """
+      Tests each scenario on the given estimator.
+
+      Returns:
+        tuple: (elapsed time, estimator name, scenario name)
+      """
+      scenario_config, sketch_estimator_config = args
+      # logging.info(f'Estimator: {sketch_estimator_config.name}'
```
kungfucraig, line 253: Delete

src/evaluations/evaluator.py
```diff
@@ -224,38 +237,94 @@ def __init__(self, evaluation_config, sketch_estimator_config_list, run_name,
     self.scenario_random_states = scenario_random_states
 
   def __call__(self):
-    self.evaluate_all()
-
-  def evaluate_all(self):
-    """Evaluate all estimators under all scenarios."""
+    self.evaluate_all_parallel()
+
+  def evaluate_all_parallel(self):
```
kungfucraig, line 242: Can you consider breaking this function up into multiple smaller methods or at least moving the very long nested method out to a private module level function?

kungfucraig, line 242: Should there be some tests added?

src/evaluations/evaluator.py
pfhgetty: Well, in this case it's actually # of processes and generally the context-switching cost of processes outweighs any benefits of having more than a single one of them per core. Of course, I don't actually need to bound it and semantically it might make more sense to simply call them workers instead of cores. That might make things less arbitrarily limiting.

src/evaluations/evaluator.py
pfhgetty: This is explained in the PR summary, but basically we need to serialize the arguments we are inputting into the process we are spawning when using multiprocessing. In the multiprocessing library, the arguments are serialized with Pickle which cannot handle some of the arguments we are inputting (namely, local functions inside of our estimators configs). Pathos is a library which has all the functionality of Python's native multiprocessing library, but uses Dill to serialize arguments instead, allowing it to serialize all of our arguments with the least amount of code change to our project.

src/evaluations/evaluator.py
pfhgetty: This function has the same output as the previous evaluate_all function, so I'm just using those tests. If I see a need for extra unit tests, I will add them in the future, but for now the API and tests are remaining the same.

pfhgetty: The nested method has been moved into a private class-level function in 7ea9926.

src/evaluations/evaluator.py
pfhgetty: Resolved in 7ea9926

src/evaluations/evaluator.py
pfhgetty: Resolved in 7ea9926

src/evaluations/evaluator.py
```diff
@@ -224,38 +237,94 @@ def __init__(self, evaluation_config, sketch_estimator_config_list, run_name,
...
+      # Update progress
+      pbar.update()
+      pbar.set_description(f'Finished {sketch_estimator_name}-{scenario_name}')
+
+    # Write performance stats
+    for time_file, elapsed_time in performance_stats.items():
       with open(time_file, 'w') as f:
-        f.write(elapsed_time)
+        f.write(str(elapsed_time))
+    pbar.close()
```
kungfucraig, line 289: Is it possible to use pbar with a context manager?

pfhgetty: Yes, would that be preferable? I tend to avoid using with statements if they wrap something relatively large

src/evaluations/run_evaluation.py
```diff
@@ -64,6 +64,10 @@
     'The name of this evaluation run.')
 flags.DEFINE_integer(
     'num_runs', None, 'The number of runs per scenario.', lower_bound=1)
+flags.DEFINE_integer(
+    'parallel_cores', 0,
```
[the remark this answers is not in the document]
pfhgetty: Resolved in 7ea9926

src/evaluations/evaluator.py
jgoodknight: It also looks, Craig, like Pathos might allow us to easily wield an army of multiple machines on different clouds but I'm not sure yet:
https://github.com/uqfoundation/pathos

src/evaluations/run_evaluation.py
```diff
@@ -64,6 +64,10 @@
     'The name of this evaluation run.')
 flags.DEFINE_integer(
     'num_runs', None, 'The number of runs per scenario.', lower_bound=1)
+flags.DEFINE_integer(
+    'workers', 0,
```
jgoodknight, line 68: should the default be the max?  This could have rough consequences for the user if they inadvertently spin up all the processes in their machine.  1 might be too much, but perhaps 2: what do you think, or would you still prefer 0?

pfhgetty: Considering that when we do run these evaluations we would want them to run as fast as possible, I don't think it's unjustified to use the maximum amount of processes that we can spin up. I see the possibility of accidentally using many fewer cores than you need as more damaging than using many more cores than you need. The consequences of using many fewer cores could be days lost while using many more cores than needed means that other people using the machine might see some slowdown in their work. Either way, it's recommended that you just set it to what you need, but the default is that the evaluations will go as fast as possible.

jgoodknight: right but I think the core of my question is: is the common case in terms of the number of times this script will be called, someone tinkering around with the evaluation script trying to do a test run, or is the common case the full shebang where the user wants things to go as fast as possible.  my suspicion is the former but could definitely be convinced otherwise

src/evaluations/evaluator.py
kungfucraig: With respect to wrapping something large, make some helper functions. The loop that starts on line 271 is a good candidate for turning into a private method called "wait_for_worker_completion".  With that out of the way the amount of code the with statement wraps will be much simpler.

Using context managers to ensure cleanup is a really good practice. Here maybe it doesn't matter that much, in the sense that I'm not sure if tqdm acquires resources. Maybe if you don't close it the pbar your terminal ends up needing to be reset? 

In retrospect though, I'm not sure taking the dependency on tqdm is worth it.

@jgoodknight WDYT?

src/evaluations/run_evaluation.py
kungfucraig, line 68: num_workers to match style of flag "num_runs" immediately above.

pfhgetty: Resolved in 2bcd89f

src/evaluations/evaluator.py
pfhgetty: Resolved in 2bcd89f

jgoodknight: Tqdm is a pretty cool addition with strongly positive usability benefits to folks running long jobs.  I'm a big fan of it and it's super common in the ml world; it's a dependency in TensorFlow if I'm not mistaken.

PR #18, opened 2020-07-18
Adds ExactMultiSet for reporting exact frequencies

matthewclegg:


src/estimators/tests/exact_set_test.py
```diff
@@ -93,6 +93,18 @@ def test_less_one_estimator_multiple_for_exact_multi_set(self):
...
+    s = ExactMultiSet()
+    s.add_ids([1,1,2,2,3,3,3])
+    e = LessOneEstimator()
+    self.assertEqual(e([s]), [1, 1, 1])
+
+  def test_less_one_estimator_no_freq1or2(self):
+    s = ExactMultiSet()
+    s.add_ids([1,1,1,2,2,2,3,3,3])
+    e = LessOneEstimator()
+    self.assertEqual(e([s]), [0, 1, 2])
```
my2457, line 106: Seems we've already decided to go with e[i]=(cardinality of id's with (f==i+1) but not with (f>=i+1))
but I think this definition would make bloom filter based solution less accurate since previously what appears to be the full cardinality is now defined as (f==1), rather than f>=1
https://github.com/world-federation-of-advertisers/cardinality_estimation_evaluation_framework/blob/freqfw2/src/estimators/bloom_filters.py#L360-L364

matthewclegg: Thank you again for pointing this out.  I have updated the code to conform to the doc.  Let me know what you think.

src/estimators/exact_set.py
```diff
@@ -84,7 +100,10 @@ def __call__(self, sketch_list):
     """Return len(sketch) - 1."""
```
my2457, line 100: Can we update the comment here to explain that we're returning cardinality short by 1 for all frequency?

matthewclegg: Done.

src/estimators/exact_set.py
```diff
@@ -53,12 +55,20 @@ def __contains__(self, x):
 
   def add(self, x):
     """Adds an id x to the sketch."""
-    self._ids.add(x)
+    self._ids[x] = self._ids.get(x, 0) + 1
 
   def ids(self):
     """Return the internal ID set."""
```
kungfucraig, line 61: Update doc string s/set/multiset ?

src/estimators/exact_set.py
```diff
@@ -41,7 +43,7 @@ def __init__(self, random_seed=None):
     """
     SketchBase.__init__(self)
     _ = random_seed
-    self._ids = set()
+    self._ids = {}
 
   def __len__(self):
     """Return the number of elements in the sketch."""
```
kungfucraig, line 49: s/number/number of unique

src/estimators/exact_set.py
```diff
@@ -53,12 +55,20 @@ def __contains__(self, x):
...
   def ids(self):
     """Return the internal ID set."""
     return self._ids
 
+  def frequency(self, x):
+    """Returns the frequency of occurrence of x."""
+    return self._ids.get(x, 0)
+
+
+class ExactSet(ExactMultiSet):
```
kungfucraig, line 69: The semantics of this inheritance are weird.

I would not expect to be able to pass as ExactSet anywhere an ExactMultiSet is required. I also would not expect "ExactSet" to have a "frequency" method.

It saves some code like this, but deriving from SketchBase, aggregating an ExactMultiSet and forwarding the methods would be cleaner.

src/estimators/exact_set.py
```diff
@@ -67,11 +77,18 @@ def __init__(self):
...
+    union = ExactMultiSet()
     for s in sketch_list:
-      union.add_ids(s.ids())
-    return [len(union)]
+      for id in s.ids():
+        union.add_ids([id] * s.frequency(id))
+    frequency_counts = collections.defaultdict(int)
+    for x in union.ids():
+      frequency_counts[union.frequency(x)] += 1
+    histogram = list(reversed(np.cumsum([frequency_counts[i] for i in range(max(frequency_counts), 0, -1)])))
```
kungfucraig, line 90: This is a little much to have in a single list comprehension. :)

src/estimators/exact_set.py
```diff
@@ -81,10 +98,13 @@ def __init__(self):
...
 
   def __call__(self, sketch_list):
-    """Return len(sketch) - 1."""
+    """Returns frequency histogram of sketch with total cardinality reduced by one."""
     e = LosslessEstimator()
     histogram = e(sketch_list)
-    return [histogram[0]-1] + [histogram[1:]]
+    if sum(histogram) == 0:
+      raise ValueError("Attempt to create a histogram with a negative value!")
+    histogram = [min(h, histogram[0]-1) for h in histogram]
```
kungfucraig, line 106: I don't understand what min(h, histogram[0]-1) is doing. I don't see how this gets you what your docstring is promising anyway.

It's possible that this "LessOneEstimator" does not make much sense in the frequency case.

Also, you don't really need the temporary variable.

src/estimators/exact_set.py
matthewclegg: Done.

src/estimators/exact_set.py
matthewclegg: Done.

src/estimators/exact_set.py
matthewclegg: I guess I subscribe to the philosophy that twice as much code equals twice as many opportunities for undetected bugs and twice as much maintenance headache.  But I agree with the point of view that if ExactSet has a frequency method, it should report something that is appropriate for a set.  Let me know if you are OK with these changes.

src/estimators/exact_set.py
matthewclegg: I've added some comments explaining what is happening.

src/estimators/exact_set.py
matthewclegg: I've added an explanation of what is happening.

src/estimators/exact_set.py
```diff
@@ -67,24 +85,47 @@ def __init__(self):
...
     histogram = e(sketch_list)
-    return [histogram[0]-1] + [histogram[1:]]
+    if sum(histogram) == 0:
+      raise ValueError("Attempt to create a histogram with a negative value!")
+    # It seems pretty clear that if the original histogram is [3, 2, 1], then
+    # a reasonable output is [2, 2, 1].  This corresponds to removing (the) one item
+    # whose frequency is 1.  But what if the histogram is [2, 2, 1]?  This represents
+    # a set having 0 items of frequency 1, 1 of frequency 2, and 1 of frequency 3.
+    # What should be removed in this case, and what should the resulting histogram
+    # look like?  I have opted for removing an item of lowest positive frequency.
```
kungfucraig, line 127: This seems too clever/complex. Maybe just subtract one from each non-zero frequency?

The whole point of this class is really just as a test utility anyway.

src/estimators/exact_set.py
kungfucraig: The list comprehension is too long/complex. A couple of temporary variables would make the comment moot too I think.

src/estimators/exact_set.py
kungfucraig: Per offline conversation we agreed to simply replace ExactSet with ExactMultiSet.

src/estimators/tests/exact_set_test.py
```diff
@@ -22,43 +22,66 @@
...
+class ExactMultiSetTest(absltest.TestCase):
 
-  def test_sketch(self):
-    s = ExactSet()
+  def test_sketch_for_exact_multi_set(self):
+    s = ExactMultiSet()
     s.add_ids([1, 2])
     self.assertLen(s, 2)
     self.assertIn(2, s)
     self.assertNotIn(3, s)
```
huangxichen1, line 37: Hi Matthew, would you mind adding error messages to these three assertions so that it is easier for debugging.

src/estimators/tests/exact_set_test.py
```diff
@@ -22,43 +22,66 @@
...
 
 from wfa_cardinality_estimation_evaluation_framework.estimators.exact_set import AddRandomElementsNoiser
-from wfa_cardinality_estimation_evaluation_framework.estimators.exact_set import ExactSet
+from wfa_cardinality_estimation_evaluation_framework.estimators.exact_set import ExactMultiSet
 from wfa_cardinality_estimation_evaluation_framework.estimators.exact_set import LessOneEstimator
 from wfa_cardinality_estimation_evaluation_framework.estimators.exact_set import LosslessEstimator
 
 
-class ExactSetTest(absltest.TestCase):
+class ExactMultiSetTest(absltest.TestCase):
```
huangxichen1, line 30: It would be nice to add a unit test for testing LosslessEstimator and LessOneEstimator estimating an empty exact multiset (an exact multiset which no IDs are added to it).

src/estimators/exact_set.py
```diff
@@ -33,18 +35,18 @@ def f(random_seed):
...
+    """Construct an ExactMultiSet sketch.
 
     Args:
       random_seed: This arg exists in order to conform to
         simulator.EstimatorConfig.sketch_factory.
     """
     SketchBase.__init__(self)
     _ = random_seed
-    self._ids = set()
+    self._ids = {}
```
jgoodknight, line 46: It doesn't matter, but this could be a collections.Counter object, which basically all it does is automatically do obj.get(key, 0) when you do obj[key] += 1

matthewclegg: I actually looked at that as a choice, but in the end I concluded it didn't really save anything.  But thanks for the suggestion!

PR #19, opened 2020-07-23
Updates DataFrame generated by simulator to contain addl columns

matthewclegg:
This pull request updates the DataFrame generated by the simulator to contain columns ESTIMATED_CARDINALTY_k, TRUE_CARDINALITY_k and RELATIVE_ERROR_k for frequencies 1,2,...,f.


src/simulations/simulator.py
```diff
@@ -134,12 +140,20 @@ def run_all_and_aggregate(self):
 
     return df, df_agg
 
+  def _extend_histogram(self, histogram, max_freq):
+    """Extends (or truncates) a frequency histogram to max_freq frequencies."""
+    if len(histogram) <= max_freq:
+      return histogram + [0] * (max_freq - len(histogram))
+    else:
+      return histogram[:max_freq]
```
huangxichen1, line 148: Here the code truncates at freq = max_freq. Does it make sense that we merge the freq categories into one when freq >= max_freq? For example, when max_freq is set to 3, and the estimator returns freq = 1, 2, 3, 4, then we merge 3 and 4 into 3+. In this way, we won't miss a certain proportion of IDs so that two different histograms can be compared with each other in terms of their distributions.

src/simulations/tests/simulator_test.py
```diff
@@ -210,8 +231,42 @@ def test_get_sketch_different_runs_different_random_state(self):
...
+        set_generator_factory=set_generator_factory,
+        sketch_estimator_config=sketch_estimator_config)
+    df, _ = sim()
+    expected_columns = ['num_sets', 'estimated_cardinality_1', 'estimated_cardinality_2',
+                        'estimated_cardinality_3', 'true_cardinality_1',
+                        'true_cardinality_2', 'true_cardinality_3', 'run_index',
+                        'relative_error_1', 'relative_error_2', 'relative_error_3']
+    expected_data = [
+      [1, 3, 2, 1, 3, 2, 1, 0, 0., 0., 0.],
+      [2, 4, 3, 2, 4, 3, 2, 0, 0., 0., 0.]
```
huangxichen1, line 264: Nit. Do we indent by two or four spaces in this context?

src/simulations/tests/simulator_test.py
```diff
@@ -99,7 +120,7 @@ def test_simulator_run_one_with_estimate_noiser(self):
     sim = get_simple_simulator(sketch_estimator_config)
     data_frame = sim.run_one()
     self.assertLen(data_frame, 1)
-    self.assertEqual(data_frame['estimated_cardinality'].iloc[0], 10)
+    self.assertEqual(data_frame['estimated_cardinality_1'].iloc[0], 10)
```
huangxichen1, line 123: I think it is my problem when I originally wrote the code that uses the hard coded column name ('estimated_cardinality') rather than using the global variable.
Matthew, do you think that it makes sense to use global variable (i.e., ESTIMATED_CARDINALITY_BASENAME) and hereafter in this file? So if we changed the name in the future, it would involve less code change and have better maintenance.

src/simulations/simulator.py
matthewclegg: Lets say the seq of id's is 1,1,1,1,1,2,2,2,2,3,3,3,4,4,5.  So the full frequency histogram is then 15,10,6,3,1.  If we truncate this at max_freq=3, it becomes 15,10,6.  None of the id's are dropped, because the the 3+ bucket counts id's have a frequency of 3 or higher.  Let me know if this makes sense.

src/simulations/tests/simulator_test.py
matthewclegg: Done

src/simulations/tests/simulator_test.py
matthewclegg: Done.

kungfucraig:
@my2457 @AaronRanAn 

We're going to submit this. Feel free to continue/start review post submission if you'd like.

PR #22, opened 2020-07-25
Add Uniform bloom and geo bloom filter

my2457:
1. Convert UnionEstimator to allow denoiser
2. Add both Uniform bloom filter and geo


src/estimators/bloom_filters.py
```diff
@@ -244,19 +248,19 @@ def _check_compatibility(cls, sketch_list):
...
-  def union_sketches(cls, sketch_list):
+  def union_sketches(self, sketch_list):
     """Exposed for testing."""
     UnionEstimator._check_compatibility(sketch_list)
-    union = copy.deepcopy(sketch_list[0])
+    sketch_list = self._denoiser(sketch_list)
+    union = sketch_list[0]
     for cur_sketch in sketch_list[1:]:
-      union.sketch = union.sketch + cur_sketch.sketch
+      union.sketch = 1 - (1 - union.sketch) * (1 - cur_sketch.sketch)
```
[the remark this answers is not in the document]
my2457: Yes!

src/estimators/bloom_filters.py
```diff
@@ -383,6 +383,21 @@ def first_moment(u):
 
     return invert_monotonic(first_moment, lower_bound)(0)
 
+  @classmethod
+  def _estimate_cardinality_geo(cls, sketch, weights):
```
kungfucraig, line 387: @shengma-1 @huangxichen1 

I didn't review this for correctness of estimator, can you guys make sure you do?

src/estimators/bloom_filters.py
```diff
@@ -396,6 +411,9 @@ def __call__(self, sketch_list):
       return [FirstMomentEstimator._estimate_cardinality_exp(union)]
     if self._method == FirstMomentEstimator.METHOD_UNIFORM:
       return [FirstMomentEstimator._estimate_cardinality_uniform(union)]
+    if self._method == FirstMomentEstimator.METHOD_GEO:
```
kungfucraig, line 414: This class is getting crowded. There is a pretty good case for breaking it up into a subclass per method. Can you add a comment about this somewhere, perhaps around line 286? Maybe the next person who adds a method will take this on.

src/estimators/bloom_filters.py
my2457: The reason I add this is to add a boundary check for denoised bloom filter to avoid the crash since in the original estimator, `1 - np.average(sketch.sketch) ` may be negative and will crash during the evaluation. I also optimized away `weights` since they're always 1. Other than this, there shouldn't be any differences

src/estimators/bloom_filters.py
my2457: Yes, agree, I'll add comments and update as a follow up

PR #23, opened 2020-07-27
Updates simulator to include shuffle distance.

matthewclegg:


src/simulations/simulator.py
```diff
@@ -28,13 +28,14 @@
...
 NUM_SETS = 'num_sets'
-
+SHUFFLE_DISTANCE = 'shuffle_distance'
 
 _SketchEstimatorConfig = collections.namedtuple(
     'EstimatorConfig', ['name', 'sketch_factory', 'estimator', 'sketch_noiser',
                         'estimate_noiser', 'max_frequency'])
 
 
+
```
huangxichen1, line 38: Two blank lines before class definition?

src/simulations/simulator.py
```diff
@@ -146,7 +147,32 @@ def _extend_histogram(self, histogram, max_freq):
...
+        of id's having a frequency of k+1 or greater.
+    Returns:
+      The shuffle distance between histogram1 and histogram2.
+    """
+    if len(histogram1) == 0 or len(histogram2) == 0:
+      return 0.0
+    counts1 = [histogram1[i] - histogram1[i+1] for i in range(len(histogram1)-1)] + [histogram1[-1]]
+    counts2 = [histogram2[i] - histogram2[i+1] for i in range(len(histogram2)-1)] + [histogram2[-1]]
+    max_freq = max(len(counts1), len(counts2))
+    freq1 = np.array(self._extend_histogram(counts1, max_freq)) / sum(counts1)
```
jgoodknight, line 172: this will never be that big, but np.sum would be faster in this function

src/simulations/simulator.py
```diff
@@ -146,7 +147,32 @@ def _extend_histogram(self, histogram, max_freq):
...
+    Args:
+      histogram1:  A frequency histogram.  histogram1[k] is the number
+        of id's having a frequency of k+1 or greater.
+      histogram2:  A frequency histogram.  histogram2[k] is the number
+        of id's having a frequency of k+1 or greater.
+    Returns:
+      The shuffle distance between histogram1 and histogram2.
+    """
+    if len(histogram1) == 0 or len(histogram2) == 0:
+      return 0.0
```
jgoodknight, line 168: should this be an error instead?

huangxichen1, line 168: Why is 0 returned? If a histogram is empty and the other is not, they should be different. Is my understanding correct?

src/simulations/simulator.py
matthewclegg: Done.

src/simulations/simulator.py
matthewclegg: Seems reasonable.

src/simulations/simulator.py
matthewclegg: Done

src/simulations/simulator.py
matthewclegg: Per Joey's comment, I have decided to raise an error if either histogram is empty.

src/simulations/simulator.py
```diff
@@ -146,7 +146,32 @@ def _extend_histogram(self, histogram, max_freq):
...
+      histogram1:  A frequency histogram.  histogram1[k] is the number
+        of id's having a frequency of k+1 or greater.
+      histogram2:  A frequency histogram.  histogram2[k] is the number
+        of id's having a frequency of k+1 or greater.
+    Returns:
+      The shuffle distance between histogram1 and histogram2.
+    """
+    assert len(histogram1), "Attempt to call _shuffle_distance with empty histogram1"
+    assert len(histogram2), "Attempt to call _shuffle_distance with empty histogram2"
+    counts1 = [histogram1[i] - histogram1[i+1] for i in range(len(histogram1)-1)] + [histogram1[-1]]
```
kungfucraig, line 168: 80 char line length please

src/simulations/simulator.py
```diff
@@ -146,7 +146,32 @@ def _extend_histogram(self, histogram, max_freq):
...
+    Args:
+      histogram1:  A frequency histogram.  histogram1[k] is the number
+        of id's having a frequency of k+1 or greater.
+      histogram2:  A frequency histogram.  histogram2[k] is the number
+        of id's having a frequency of k+1 or greater.
+    Returns:
+      The shuffle distance between histogram1 and histogram2.
+    """
+    assert len(histogram1), "Attempt to call _shuffle_distance with empty histogram1"
+    assert len(histogram2), "Attempt to call _shuffle_distance with empty histogram2"
```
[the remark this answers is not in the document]
matthewclegg: Hi Sheng, Thank you for the comments!  I am happy to make this change, but may I ask why?  In the current implementation, the shorted histogram is extended by appending zeros to it.  To me this seems like a reasonable and consistent interpretation.

src/simulations/simulator.py
```diff
@@ -146,7 +146,32 @@ def _extend_histogram(self, histogram, max_freq):
       return histogram + [0] * (max_freq - len(histogram))
     else:
       return histogram[:max_freq] 
-      
+
+  def _shuffle_distance(self, histogram1, histogram2):
+    """Computes shuffle distance between two histograms.
```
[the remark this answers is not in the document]
matthewclegg: The current behavior is to always normalize the probability distributions.  Is there a use case for computing the shuffle distance of non-normalized distributions?

src/simulations/simulator.py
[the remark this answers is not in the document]
matthewclegg: I meant to write, "the shorter histogram is extended..."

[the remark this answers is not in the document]
kungfucraig: +1 @matthewclegg

PR #24, opened 2020-07-28
Add frequency set generator of homogeneous user activities within a publisher.

huangxichen1:
The scenario assumes that the users within a publisher’s campaign have the same uniform Poisson frequency distribution. Also frequency capping is supported.

### DIFF
--- a/src/simulations/set_generator.py
+++ b/src/simulations/set_generator.py
@@ -115,14 +115,14 @@ class IndependentSetGenerator(SetGeneratorBase):
 
     return f
 
-  def __init__(self, universe_size, set_size_list, random_state):
+  def __init__(self, universe_size, set_sizes, random_state):
     self.universe_size = universe_size
     self.union_ids = set()
     self.random_state = random_state
-    self.set_size_list = set_size_list
+    self.set_sizes = set_sizes
 
   def __iter__(self):
-    for set_size in self.set_size_list:
+    for set_size in self.set_sizes:
       set_ids = choice_fast(self.universe_size, set_size, self.random_state)
       self.union_ids = self.union_ids.union(set_ids)
       yield set_ids
@@ -158,7 +158,7 @@ class ExponentialBowSetGenerator(SetGeneratorBase):
     return f
 
   def __init__(self, user_activity_association, universe_size,
-               set_size_list, random_state):
+               set_sizes, random_state):
     """Initialize an Exponential Bow set generator.
 
     Args:
@@ -173,7 +173,7 @@ class ExponentialBowSetGenerator(SetGeneratorBase):
         user_activity_association are not supported.
       universe_size: An integer value that specifies the size of the whole
         universe from which the ids will be sampled.
-      set_size_list: An iterator or a list containing the size of each set.
+      set_sizes: An iterator or a list containing the size of each set.
       random_state: A np.random.RandomState object.
 
     Raises:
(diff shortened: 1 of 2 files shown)


src/simulations/set_generator.py
```diff
@@ -97,14 +115,14 @@ def f(random_state):
 
     return f
 
-  def __init__(self, universe_size, set_sizes, random_state):
+  def __init__(self, universe_size, set_size_list, random_state):
```
kungfucraig, line 118: I'd continue to call this "set_sizes" since we generally expect an iterable not a list per se.

src/simulations/set_generator.py
```diff
@@ -486,3 +506,87 @@ def __iter__(self):
...
+      set_ids = self.choice(
+          self.universe_size, set_size, self.random_state)
+      freq_per_id = self.random_state.poisson(
+          lam=freq_rate, size=len(set_ids)) + 1
+      if self.freq_cap is not None:
+        freq_per_id = np.minimum(freq_per_id, self.freq_cap)
+      multiset_ids = []
+      for i, freq in zip(set_ids, freq_per_id):
+        multiset_ids += [i] * freq
+      yield multiset_ids
```
kungfucraig, line 591: Should we shuffle the multiset before returning? I think this will have an effect on how same-key-aggregator behaves. This could be worth discussing in more detail offline.

src/simulations/set_generator.py
```diff
@@ -51,14 +51,32 @@ def __iter__(self):
...
+
 class SetGeneratorBase(object):
   """Base object for generating test sets."""
 
   def __next__(self):
     raise NotImplementedError()
 
   @classmethod
-  def get_generator_factory(cls):
+  def get_generator_factory_with_num_and_size(cls):
```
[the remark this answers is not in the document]
huangxichen1: Acknowledged.

src/simulations/set_generator.py
```diff
@@ -486,3 +506,87 @@ def __iter__(self):
     for i in self.set_indices:
       yield set_ids_list[i]
     return self
+
+
+class HomogeneousMultiSetGenerator(SetGeneratorBase):
+  """Homogeneous multiset generator.
```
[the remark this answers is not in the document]
huangxichen1: Done.

src/simulations/set_generator.py
```diff
@@ -486,3 +506,87 @@ def __iter__(self):
...
+                                               set_size_list, freq_rate_list,
+                                               freq_cap):
+
+    def f(random_state):
+      return cls(universe_size, set_size_list, freq_rate_list, random_state,
+                 freq_cap)
+
+    return f
+
+  def __init__(self, universe_size, set_size_list, freq_rate_list, random_state,
```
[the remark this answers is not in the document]
huangxichen1: Done.

src/simulations/set_generator.py
huangxichen1: Done.

src/simulations/set_generator.py
huangxichen1: Done.

src/simulations/tests/set_generator_test.py
```diff
@@ -414,6 +432,62 @@ def test_sequentially_correlated_all_previous_generator_raise_error(self):
...
+    set_sizes = [1, 2]
+    freq_rate_list = [5, 1]
+    rs = FakeRandomState()
+    gen = set_generator.HomogeneousMultiSetGenerator(
+        universe_size=4,
+        set_sizes=set_sizes,
+        freq_rate_list=freq_rate_list,
+        freq_cap=3,
+        random_state=rs)
+    expected_multiset_ids_list = [[0, 0, 0], [1, 1, 0, 0]]
```
my2457, line 449: `expected_multiset_ids_list` is parameterized, so we don't need to hardcode it? My understanding is that we need to pass `freq_cap` & `expected_multiset_ids_list` to `HomogeneousMultiSetGenerator` and expect the value to be the same in parameterized params?

huangxichen1: Thanks for finding this! I've modified the code. PTAL.

PR #25, opened 2020-07-29
Adds FrequencyEstimatorEvaluationAnalyzer

matthewclegg:
Adds the FrequencyEstimatorEvaluationAnalyzer.  A later pull request will add the frequency histogram graphs mentioned in the design doc.  The purpose of this update is to facilitate building an end-to-end test of the frequency evaluation framework.

### DIFF
--- a/src/evaluations/report_generator.py
+++ b/src/evaluations/report_generator.py
@@ -60,6 +60,8 @@ MESSAGE_HTML_TEMPLATE = """
       </body>
     </html>"""
 
+# Number of digits that should be shown when reporting the relative error
+RELATIVE_ERROR_FORMAT_ACCURACY = 4
 
 class ReportGenerator(object):
   """Generate HTML report for the cardinality estimator evaluation."""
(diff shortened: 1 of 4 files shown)


src/evaluations/report_generator.py
```diff
@@ -155,9 +155,9 @@ def _format_num_estimable_sets_df(df):
       df['num_estimable_sets_cell'] = (
           df[analyzer.NUM_ESTIMABLE_SETS].astype('str')
           + '<br>relative_error: mean='
-          + df[simulator.RELATIVE_ERROR + '_mean'].round(4).astype('str')
+          + df[simulator.RELATIVE_ERROR_BASENAME + '1_mean'].round(4).astype('str')
```
jgoodknight, line 158: I know you didn't write this code, but I do wonder if the precision of the rounding should be a flag or at least a constant defined at top?

matthewclegg: Even though every CS 101 class teaches that one should never have constants in the code as above, this is one case where I think the utility of defining a manifest constant is debatable.  Nonetheless, I have made the change you requested.

src/evaluations/tests/analyzer_test.py
```diff
@@ -163,13 +163,13 @@ def test_get_relative_error_stats_of_num_of_estimable_sets(self):
         [analyzer.ERROR_MARGIN_NAME, analyzer.PROPORTION_OF_RUNS_NAME,
          analyzer.SKETCH_ESTIMATOR_NAME, analyzer.SCENARIO_NAME,
          analyzer.NUM_ESTIMABLE_SETS,
-         simulator.RELATIVE_ERROR + '_mean',
-         simulator.RELATIVE_ERROR + '_std'],
+         simulator.RELATIVE_ERROR_BASENAME + '1_mean',
```
kungfucraig, line 166: Given you use "1_mean" twice, it might be nice to store in a variable. For future proofing sake.

PR #28, opened 2020-07-31
Add evaluator, analyzer and report_generator to the interoperability …

huangxichen1:
…test.


tests/interoperability_test.py
```diff
@@ -389,5 +415,36 @@ def test_with_seq_corr_one_set_generator_non_noised_different_sizes(self):
     self.simulate_with_set_generator(set_generator_factory,
                                      self.name_to_noised_estimator_config)
 
+  def test_run_evaluation_works(self):
```
matthewclegg, line 418: This unit test name seems a little opaque.  According to Google's guidelines, "Test names should include the condition being tested and the expected outcome."  Could possibly be addressed by adding a comment explaining further, but probably the best practice would be to make the name more descriptive.

huangxichen1: I changed the test name and add some docstring to explain more details. PTAL. Thanks!

huangxichen1:
> Try rebase? Test failed on my machine probably due to some preexisting merge issues

Done. The test can run now. Yet I find there is another bug needed to be fixed in the other files. I will create a separate PR to fix this.

PR #30, opened 2020-08-01
Correct report link in the readme file.

huangxichen1:


results/README.md
```diff
@@ -8,4 +8,4 @@ record here.
 
 | directory | commit No. | evaluation | estimators | report link |
 |-|-|-|-|-|
```
shengma-1, line 10: Can you add the instruction of viewing html directly from github?

huangxichen1: Ack. I will do in a separate PR.

PR #31, opened 2020-08-04
Updates run_evaluation.py with capability to do frequency evaluations.

matthewclegg:
This update is the (next to) last step in creating the end-to-end test of the frequency evaluator.  I decided to add a command line flag, --frequency_analysis, which generates a report that analyzes performance by frequency.  In this way, it will be possible to generate both cardinality reports and frequency reports for estimators that are capable of reporting frequency.  Tomorrow, I will add a sample config, which will be the final step in demonstrating the end-to-end functionality.


src/evaluations/run_evaluation.py
```diff
@@ -80,6 +80,8 @@
                      'The widths of the boxplot in inches.')
 flags.DEFINE_integer('boxplot_size_height_inch', 6,
                      'The widths of the boxplot in inches.')
+flags.DEFINE_bool('frequency_analysis', False,
```
kungfucraig, line 83: I think a flag indicating the type of analysis with a default of "cardinality" would be cleaner.

matthewclegg: Done.

my2457:
So we will generate frequency report with flag `--analysis_type="frequency"`, correct?

src/evaluations/report_generator.py
```diff
@@ -140,27 +140,29 @@ def add_parsed_sketch_estimator_name_cols(cls, df, sketch_estimator_name_col):
...
       stats.
     """
-    def _format_num_estimable_sets_df(df):
+    def _format_num_estimable_sets_df(df, error_metric_column):
       df = df.copy()
       df['num_estimable_sets_cell'] = (
           df[analyzer.NUM_ESTIMABLE_SETS].astype('str')
           + '<br>relative_error: mean='
-          + df[simulator.RELATIVE_ERROR_BASENAME + '1_mean'].round(
+          + df[error_metric_column + '_mean'].round(
```
my2457, line 162: curious why do we need appended `1` for reach, but not for frequency?

matthewclegg: We now compute relative error by frequency, so we have relative_error_1, relative_error_2, ..., where relative_error_k represents the estimation error for frequency k or more.  Thus, in this new definition, relative_error_1 has the same meaning as relative_error had in the previous version of the code.  For frequency, we use the shuffle_distance metric rather than the relative_error metric.  Shuffle_distance is something like the average relative error over all frequencies (up to some maximum frequency).  So, it is not frequency-specific.  I hope this explanation helps!

PR #32, opened 2020-08-04
smoke test results for uniform and geo bloom filter

my2457:
smoke test result for uniform and geo bloom filter


huangxichen1:
Thanks, Miao!

Just curious, why do you use different length (in magnitude) for the uniform BF and the Geometric BF?

my2457:
> Thanks, Miao!
> 
> Just curious, why do you use different length (in magnitude) for the uniform BF and the Geometric BF?

Regular bloom filter needs lots of more bits but geometric doesn't really (similar to log/exp)

my2457:
> Should the same bit size be used for BF and geo-adbf for reasonable comparison?
> We use 10^5 for exp-/log-adbf.

I think this is fine as I found that it has more to do with the geo parameter rather than the bit size.

PR #33, opened 2020-08-05
Fix bug in the report_generator and the interoperability test by usin…

huangxichen1:
…g the global variable.


kungfucraig:
@AaronRanAn @my2457 

We're submitting a quick bug fix. Feel free to review after submission if you'd like.

PR #34, opened 2020-08-05
Modify the complete evaluation configs and add additional tests.

huangxichen1:
1. Correct the complete evaluation function, so it should contain a list of scenario configs.
2. Add unit test for 1.
3. Modify the mock patch for testing if the complete evaluation function uses all the scenarios. Now it should work with both
(1) python run_tests.py --folders=src/evaluations/data/
(2) python src/evaluations/data/tests/evaluation_configs_test.py
The previous PR only works with (2).


kungfucraig:
@AaronRanAn @my2457 

We're going to submit this small bugfix as well.

PR #35, opened 2020-08-06
Fix sequentially correlated set generator.

huangxichen1:
1. Sequentially correlated set generator will run into error if the new set is smaller than the overlap set size. This PR fix this bug by capping the size of the overlap set if it is larger than the new set. Also add a unit test for this feature.
2. The sequentially correlated set generator does not need the universe size. Remove this argument in order to reduce potential bug when creating the evaluation configuration. Also modified the interoperability test and the evaluation configs and its tests.

### DIFF
--- a/src/simulations/set_generator.py
+++ b/src/simulations/set_generator.py
@@ -397,13 +397,13 @@ class _SequentiallyCorrelatedThePreviousSetGenerator(SetGeneratorBase):
 
     Every newly generated set has some overlap with THE previously generated
     set. The overlap is determined by the previous set size multiplied by the
-    shared proportion. If the shared proportion is too large, will cap it by
-    the current set size.
+    shared proportion. If the previous set is not large enough, then will the
+    overlap will use the previous set itself.
 
     Args:
       shared_prop: a number between 0 and 1, indicating the proportion of ids of
         the current set coming from the previously generated set. I.e.,
-        overlap size divided by the previous set size.
+        overlap size divided by the current set size.
       set_size_list: a list of the integer numbers representing the set size of
         the sets.
       random_state: a numpy.random.RandomState instance.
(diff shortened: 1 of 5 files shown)


src/simulations/set_generator.py
```diff
@@ -331,29 +332,33 @@ def __iter__(self):
...
+
 class _SequentiallyCorrelatedAllPreviousSetGenerator(SetGeneratorBase):
   """Generator for sequentailly correlated sets.
 
   Each set has some overlap with the union of previously generated campaigns.
   """
 
-  def __init__(self, universe_size, shared_prop, set_size_list, random_state):
-    self.universe_size = universe_size
+  def __init__(self, shared_prop, set_size_list, random_state):
```
kungfucraig, line 342: Can you document the input params?

src/simulations/set_generator.py
```diff
@@ -331,29 +332,33 @@ def __iter__(self):
...
-                                 for set_size in self.set_size_list[:-1]]
-                              + [0])
-    # self.overlap_size_list[i] is the overlap of set.set_size_list[i]
-    # with the previous union, so in particular, self.overlap_size_list[0] = 0
-    total_ids_size = int(sum(self.set_size_list) - sum(self.overlap_size_list))
-    self.ids_pool = choice_fast(universe_size, total_ids_size, self.random_state)
+    self.num_sets = len(set_size_list)
+    self.overlap_size_list = [0]
+    total_ids_size = set_size_list[0]
+    for i in range(self.num_sets - 1):
```
kungfucraig, line 349: Can you add comments about what you're doing here. Basically just transfer the description from the CL. I think it's not straightforward why this code exists.

src/simulations/tests/set_generator_test.py
```diff
@@ -407,13 +412,30 @@ def test_sequentially_correlated_all_previous_generator_raise_error(self):
...
 
+  @parameterized.parameters(
+      (set_generator.CORRELATED_SETS_ALL,),
+      (set_generator.CORRELATED_SETS_ONE,))
+  def test_sequentially_correlated_generator_overlap_size_larger_than_set_size(
+      self, correlation_type):
+    sc_gen = set_generator.SequentiallyCorrelatedSetGenerator(
+        order=set_generator.ORDER_ORIGINAL, correlated_sets=correlation_type,
+        shared_prop=0.5, set_sizes=[10, 1],
+        random_state=np.random.RandomState())
```
kungfucraig, line 430: You sure you don't want a constant seed here?

src/simulations/set_generator.py
huangxichen1: Done.

src/simulations/tests/set_generator_test.py
huangxichen1: Added a constant seed.

src/simulations/set_generator.py
huangxichen1: Done.

jiayu-google:
Thank Xichen for further making the codes consistent!

Now the shared proportion means overlap / **current_set_size**, for both the "correlated with all" and "correlated with THE previous" sub scenarios.

PR #36, opened 2020-08-06
End-to-end test

matthewclegg:
This is the end-to-end test of the recent changes to support frequency evaluation.  To run this test, use the following command line:

python3 src/evaluations/run_evaluation.py \
  --evaluation_out_dir="/tmp/evf/output/debug" \
  --analysis_out_dir="/tmp/evf/output/debug" \
  --report_out_dir="/tmp/evf/output/debug" \
  --evaluation_config="frequency_end_to_end_test" \
  --sketch_estimator_configs="exact_multi_set-10000-NA-NA" \
  --evaluation_run_name="simple_run" \
  --analysis_type=frequency \
  --num_workers=1 \
  --num_runs=3

Please note that many (most) of the changes made to evaluation_configs.py are temporary.  When the full smoke test is implemented in two weeks, the end-to-end test will be replaced.


src/evaluations/data/evaluation_configs.py
```diff
@@ -379,17 +386,50 @@ def _complete_test_with_selected_parameters(
...
+        set_generator_factory=(
+          set_generator.HomogeneousMultiSetGenerator
+          .get_generator_factory_with_num_and_size(
+            universe_size=universe_size,
+            num_sets=num_sets,
+            set_size=set_size,
+            freq_rate_list = freq_rate_list,
+            freq_cap=freq_cap)))]
+  )
+
```
huangxichen1, line 415: nit. 2 lines before function calls.

src/evaluations/data/evaluation_configs.py
```diff
@@ -379,17 +386,50 @@ def _complete_test_with_selected_parameters(
...
-EVALUATION_CONFIGS_TUPLE = (
+def _frequency_end_to_end_test(num_runs=NUM_RUNS_VALUE):
+  """EvaluationConfig of end-to-end test of frequency evaluation code."""
+  num_sets=3
+  universe_size=10000
+  set_size=5000
+  freq_rate_list=[1,2,3]
+  freq_cap=5
+  return EvaluationConfig(
+    name='frequency_end_to_end_test',
```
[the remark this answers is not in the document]
matthewclegg: Thanks for pointing this out.

src/evaluations/data/evaluation_configs.py
matthewclegg: And this too.

src/evaluations/data/evaluation_configs.py
```diff
@@ -456,17 +498,49 @@ def _complete_test_with_selected_parameters(
...
+      VECTOR_OF_COUNTS_4096_LN3_SEQUENTIAL,
+      VECTOR_OF_COUNTS_4096_INFTY_SEQUENTIAL)
+
+def _generate_frequency_estimator_configs():
+  return (
+      SketchEstimatorConfig(
+          name='exact_multi_set-10000-NA-NA',
+          sketch_factory=exact_set.ExactMultiSet.get_sketch_factory(),
+          estimator=exact_set.LosslessEstimator(),
+          max_frequency=FLAGS.max_frequency),
```
kungfucraig, line 518: Can this be passed in?

src/evaluations/data/evaluation_configs.py
```diff
@@ -14,14 +14,21 @@
...
 from wfa_cardinality_estimation_evaluation_framework.estimators import liquid_legions
 from wfa_cardinality_estimation_evaluation_framework.estimators import vector_of_counts
 from wfa_cardinality_estimation_evaluation_framework.evaluations.configs import EvaluationConfig
 from wfa_cardinality_estimation_evaluation_framework.evaluations.configs import ScenarioConfig
 from wfa_cardinality_estimation_evaluation_framework.simulations import set_generator
 from wfa_cardinality_estimation_evaluation_framework.simulations.simulator import SketchEstimatorConfig
 
+FLAGS = flags.FLAGS
+
+flags.DEFINE_integer('max_frequency', 10, 'Maximum frequency to be analyzed.')
```
kungfucraig, line 30: I'd prefer if we avoid declaring flags in files that do not have main functions.

src/evaluations/data/evaluation_configs.py
```diff
@@ -379,17 +386,52 @@ def _complete_test_with_selected_parameters(
...
+def get_evaluation_config(config_name):
+  """Returns the evaluation config with the specified config_name."""
+  configs = _generate_evaluation_configs()
+  config = [c for c in configs if c().name == config_name]
+  if not config:
+    valid_config_names = [c().name for c in configs]
+    raise ValueError("Invalid evaluation config: {}\n"
+                     "Valid choices are as follows: {}".format(
+                     config_name, ','.join(valid_config_names)))
+  return config[0]
```
kungfucraig, line 434: What if there is more than one with same name? Maybe add an error check for that case?

matthewclegg: I suspect that you had in mind that I check whether len(config) == 1.  However, my opinion is that this would result in a strange behavior.  Arguably, it is an error to define two configs with the same name.  The detection of this error should be independent of the parameters passed on the command line.  So, I added some code to the top of this function to check if any of the config names are repeated. LMKWYT.

kungfucraig: As long as we can ensure there's exactly one thing here, that works for me.

PR #38, opened 2020-08-11
Add epsilon arg to the surreal denoiser.

huangxichen1:
Add epsilon argument to the surreal denoiser and related unit tests, in order to support more comprehensive evaluations of ADBF under different epsilons.

### DIFF
--- a/src/estimators/bloom_filters.py
+++ b/src/estimators/bloom_filters.py
@@ -167,7 +167,8 @@ class GeometricBloomFilter(AnyDistributionBloomFilter):
     super().__init__(
         any_sketch.SketchConfig([
             any_sketch.IndexSpecification(
-                any_sketch.GeometricDistribution(length, probability), "geometric")
+                any_sketch.GeometricDistribution(length, probability),
+                "geometric")
         ], num_hashes=1, value_functions=[any_sketch.BitwiseOrFunction()]),
         random_seed)
 
@@ -435,6 +436,19 @@ class FixedProbabilityBitFlipNoiser(SketchNoiserBase):
     return new_filter
 
 
+def get_probability_of_flip(epsilon, num_hashes):
+  """Get the flipping probability from the privacy epsilon.
+
+  Args:
+    epsilon: the differential privacy parameter.
+    num_hashes: the number of hash functions used by the bloom filter.
+
+  Returns:
+    The flipping probability.
+  """
+  return 1 / (1 + math.exp(epsilon / num_hashes))
+
+
 class BlipNoiser(SketchNoiserBase):
   """This class applies "Blip" noise to a BloomFilter.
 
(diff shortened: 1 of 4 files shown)


src/estimators/bloom_filters.py
```diff
@@ -489,14 +501,33 @@ def __call__(self, sketch_list):
 class SurrealDenoiser(DenoiserBase):
   """A closed form denoiser for a list of Any Distribution Bloom Filter."""
 
-  def __init__(self, probability=None, flip_one_probability=None,
+  def __init__(self, epsilon=None, probability=None, flip_one_probability=None,
```
kungfucraig, line 504: Is it possible to remove the probability arg? I realize it will take some refactoring, but leaving it around causes some ugliness below. 

If we must keep it creating a classmethod that takes an epsilon and creates a SurrealDenoiser would be cleaner. 

Complex relationships between args is not good.

src/estimators/bloom_filters.py
```diff
@@ -501,14 +501,12 @@ def __call__(self, sketch_list):
 class SurrealDenoiser(DenoiserBase):
   """A closed form denoiser for a list of Any Distribution Bloom Filter."""
 
-  def __init__(self, epsilon=None, probability=None, flip_one_probability=None,
+  def __init__(self, epsilon=None, flip_one_probability=None,
                flip_zero_probability=None):
```
kungfucraig, line 505: Are "flip_zero" and "flip_one" still necessary?

src/evaluations/data/evaluation_configs.py
```diff
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 """Evaluation configurations."""
+import math
```
my2457, line 15: possible to just use np types (e.g. `np.log(3)` as below to be consistent for both setup?)

src/estimators/bloom_filters.py
```diff
@@ -435,6 +436,19 @@ def __call__(self, bloom_filter):
...
+  """Get the flipping probability from the privacy epsilon.
+
+  Args:
+    epsilon: the differential privacy parameter.
+    num_hashes: the number of hash functions used by the bloom filter.
+
+  Returns:
+    The flipping probability.
+  """
+  return 1 / (1 + math.exp(epsilon / num_hashes))
```
shengma-1, line 449: Maybe, add a comment to explain: The per bucket flipping probability should factor in the id sensitivity.  Multiple hashes assign an id to num_hashes bit buckets, and thus increase the sensitivity of an id to num_hashes.

src/estimators/bloom_filters.py
huangxichen1: After discussing with Jiayu, we don't need flip_x_probability now. I remove it in the new commit.

src/evaluations/data/evaluation_configs.py
huangxichen1: Thanks for noticing that! I convert all the np.log(3) to math.log(3) so as to save computation cost (though not that large).

src/estimators/bloom_filters.py
huangxichen1: Ack.

PR #39, opened 2020-08-12
Uakyol/stratified sketch init

uakyol:
Hi Matthew,

I would appreciate a review, thanks

### DIFF
--- a/examples/basic_comparison.py
+++ b/examples/basic_comparison.py
@@ -52,7 +52,7 @@ flags.DEFINE_integer('exponential_bloom_filter_decay_rate', 10,
                      'The decay rate in exponential bloom filter')
 flags.DEFINE_integer('num_bloom_filter_hashes', 3,
                      'The number of hashes for the bloom filter to use')
-flags.DEFINE_integer('max_frequency', 2, 'Maximum frequency to be analyzed.')
+flags.DEFINE_integer('max_frequency', 1, 'Maximum frequency to be analyzed.')
 flags.DEFINE_float('geometric_bloom_filter_probability', 0.0015,
                    'probability of geometric distribution')
 
@@ -107,7 +107,7 @@ def main(argv):
       estimator=LosslessEstimator())
 
   estimator_config_stratified = SketchEstimatorConfig(
-      name='stratified',
+      name='exact-stratified',
       max_frequency=FLAGS.max_frequency,
       sketch_factory=StratifiedSketch.get_sketch_factory(
           max_freq=FLAGS.max_frequency,
@@ -136,7 +136,7 @@ def main(argv):
       'exact_set': estimator_config_exact,
       'hll++': estimator_config_hll,
       'vector_of_counts': estimator_config_voc,
-      'stratified_sketch': estimator_config_stratified,
+      estimator_config_stratified.name: estimator_config_stratified,
   }
   set_generator_factory = (
       set_generator.IndependentSetGenerator
--- a/examples/basic_comparison.py
+++ b/examples/basic_comparison.py
@@ -52,8 +52,7 @@ flags.DEFINE_integer('exponential_bloom_filter_decay_rate', 10,
                      'The decay rate in exponential bloom filter')
 flags.DEFINE_integer('num_bloom_filter_hashes', 3,
                      'The number of hashes for the bloom filter to use')
-flags.DEFINE_integer('stratified_max_freq', 3,
-                     'Maximum frequency for stratified sketch')
+flags.DEFINE_integer('max_frequency', 2, 'Maximum frequency to be analyzed.')
 flags.DEFINE_float('geometric_bloom_filter_probability', 0.0015,
                    'probability of geometric distribution')
 
@@ -109,9 +108,9 @@ def main(argv):
 
   estimator_config_stratified = SketchEstimatorConfig(
       name='stratified',
-      max_frequency=FLAGS.stratified_max_freq,
+      max_frequency=FLAGS.max_frequency,
       sketch_factory=StratifiedSketch.get_sketch_factory(
-          max_freq=FLAGS.stratified_max_freq,
+          max_freq=FLAGS.max_frequency,
           cardinality_sketch_factory=ExactMultiSet.get_sketch_factory()),
       estimator=stratified_sketch.SequentialEstimator(
           sketch_operator=ExactSetOperator(),
--- a/examples/basic_comparison.py
+++ b/examples/basic_comparison.py
@@ -16,6 +16,7 @@ from absl import app
 from absl import flags
 import numpy as np
 
+from wfa_cardinality_estimation_evaluation_framework.estimators import stratified_sketch
 from wfa_cardinality_estimation_evaluation_framework.estimators.bloom_filters import BloomFilter
 from wfa_cardinality_estimation_evaluation_framework.estimators.bloom_filters import ExponentialBloomFilter
 from wfa_cardinality_estimation_evaluation_framework.estimators.bloom_filters import FirstMomentEstimator
@@ -28,6 +29,8 @@ from wfa_cardinality_estimation_evaluation_framework.estimators.exact_set import
 from wfa_cardinality_estimation_evaluation_framework.estimators.exact_set import LosslessEstimator
 from wfa_cardinality_estimation_evaluation_framework.estimators.hyper_log_log import HllCardinality
 from wfa_cardinality_estimation_evaluation_framework.estimators.hyper_log_log import HyperLogLogPlusPlus
+from wfa_cardinality_estimation_evaluation_framework.estimators.stratified_sketch import StratifiedSketch
+from wfa_cardinality_estimation_evaluation_framework.estimators.stratified_sketch import ExactSetOperator
 from wfa_cardinality_estimation_evaluation_framework.estimators.vector_of_counts import SequentialEstimator
 from wfa_cardinality_estimation_evaluation_framework.estimators.vector_of_counts import VectorOfCounts
 from wfa_cardinality_estimation_evaluation_framework.simulations import set_generator
@@ -49,8 +52,11 @@ flags.DEFINE_integer('exponential_bloom_filter_decay_rate', 10,
                      'The decay rate in exponential bloom filter')
 flags.DEFINE_integer('num_bloom_filter_hashes', 3,
                      'The number of hashes for the bloom filter to use')
+flags.DEFINE_integer('stratified_max_freq', 3,
+                     'Maximum frequency for stratified sketch')
 flags.DEFINE_float('geometric_bloom_filter_probability', 0.0015,
-                     'probability of geometric distribution')
+                   'probability of geometric distribution')
+
 
 def main(argv):
   if len(argv) > 1:
@@ -101,6 +107,16 @@ def main(argv):
       sketch_factory=ExactMultiSet.get_sketch_factory(),
       estimator=LosslessEstimator())
 
+  estimator_config_stratified = SketchEstimatorConfig(
+      name='stratified',
+      max_frequency=FLAGS.stratified_max_freq,
+      sketch_factory=StratifiedSketch.get_sketch_factory(
+          max_freq=FLAGS.stratified_max_freq,
+          cardinality_sketch_factory=ExactMultiSet.get_sketch_factory()),
+      estimator=stratified_sketch.SequentialEstimator(
+          sketch_operator=ExactSetOperator(),
+          cardinality_estimator=LosslessEstimator()))
+
   estimator_config_list = [
       estimator_config_bloom_filter,
       estimator_config_logarithmic_bloom_filter,
@@ -109,6 +125,7 @@ def main(argv):
       estimator_config_exact,
       estimator_config_hll,
       estimator_config_voc,
+      estimator_config_stratified,
   ]
 
   name_to_estimator_config = {
@@ -120,10 +137,11 @@ def main(argv):
       'exact_set': estimator_config_exact,
       'hll++': estimator_config_hll,
       'vector_of_counts': estimator_config_voc,
+      'stratified_sketch': estimator_config_stratified,
   }
   set_generator_factory = (
-      set_generator.IndependentSetGenerator.
-      get_generator_factory_with_num_and_size(
+      set_generator.IndependentSetGenerator
+      .get_generator_factory_with_num_and_size(
           universe_size=FLAGS.universe_size,
           num_sets=FLAGS.number_of_sets,
           set_size=FLAGS.set_size))
--- a/src/estimators/exact_set.py
+++ b/src/estimators/exact_set.py
@@ -53,6 +53,11 @@ class ExactMultiSet(SketchBase):
     """Return true if x is contained in the sketch."""
     return x in self._ids
 
+  def add_ids(self, x):
+    """Adds a list of ids to the sketch."""
+    for item_id in x:
+      self.add(item_id)
+
   def add(self, x):
     """Adds an id x to the sketch."""
     self._ids[x] = self._ids.get(x, 0) + 1
--- a/src/estimators/exact_set.py
+++ b/src/estimators/exact_set.py
@@ -53,11 +53,6 @@ class ExactMultiSet(SketchBase):
     """Return true if x is contained in the sketch."""
     return x in self._ids
 
-  def add_ids(self, x):
-    """Adds a list of ids to the sketch."""
-    for item_id in x:
-      self.add(item_id)
-
   def add(self, x):
     """Adds an id x to the sketch."""
     self._ids[x] = self._ids.get(x, 0) + 1
--- a/src/estimators/stratified_sketch.py
+++ b/src/estimators/stratified_sketch.py
@@ -164,12 +164,17 @@ class PairwiseEstimator(EstimatorBase):
     assert isinstance(this, StratifiedSketch)
     this.assert_compatible(that)
     max_freq = this.max_freq
+    max_key = str(max_freq) + '+'
     merged_sketch = copy.deepcopy(this)
+
     this_one_plus = this.cardinality_sketch_factory(0)
     that_one_plus = that.cardinality_sketch_factory(0)
     for k in range(1, max_freq):
       this_one_plus = self.sketch_union(this_one_plus, this.sketches[k])
       that_one_plus = self.sketch_union(that_one_plus, that.sketches[k])
+
+    this_one_plus = self.sketch_union(this_one_plus, this.sketches[max_key])
+    that_one_plus = self.sketch_union(that_one_plus, that.sketches[max_key])
 
     for k in range(1, max_freq):
       # Calculate A(k) & B(0) = A(k) - (A(k) & B(1+))
@@ -192,13 +197,13 @@ class PairwiseEstimator(EstimatorBase):
       merged_sketch.sketches[k] = merged
 
     # Calculate Merged(max_freq)
-    max_key = str(max_freq) + '+'
     merged = this.sketches[max_key]
     rest = that_one_plus
     for k in range(1, max_freq):
       merged = self.sketch_union(
           merged, self.sketch_intersection(this.sketches[max_freq - k], rest))
       rest = self.sketch_difference(rest, that.sketches[k])
+
     merged = self.sketch_union(
         merged,
         self.sketch_difference(
--- a/src/estimators/stratified_sketch.py
+++ b/src/estimators/stratified_sketch.py
@@ -15,6 +15,7 @@
 import copy
 import functools
 from wfa_cardinality_estimation_evaluation_framework.estimators.base import SketchBase
+from wfa_cardinality_estimation_evaluation_framework.estimators.base import EstimatorBase
 from wfa_cardinality_estimation_evaluation_framework.estimators.exact_set import ExactMultiSet
 
 ONE_PLUS = '1+'
@@ -121,7 +122,7 @@ class StratifiedSketch(object):
         f'{self.max_freq} != {other.max_freq}')
 
 
-class PairwiseEstimator(object):
+class PairwiseEstimator(EstimatorBase):
   """Merge and estimate two StratifiedSketch."""
 
   def __init__(self, sketch_operator, cardinality_estimator):
@@ -199,12 +200,9 @@ class PairwiseEstimator(object):
     merged_sketch.sketches[max_freq] = merged
 
     # Calculate Merged(1+)
-    merged_one_plus = None
-    for k in range(max_freq):
-      merged_one_plus = self.sketch_union(merged_one_plus,
-                                          merged_sketch.sketches[k + 1])
+    merged_one_plus = self.sketch_union(this.sketches[ONE_PLUS],
+                                        that.sketches[ONE_PLUS])
     merged_sketch.sketches[ONE_PLUS] = merged_one_plus
-
     return merged_sketch
 
   def estimate_cardinality(self, stratified_sketch):
@@ -223,7 +221,7 @@ class PairwiseEstimator(object):
     return result
 
 
-class SequentialEstimator(object):
+class SequentialEstimator(EstimatorBase):
   """Sequential frequency estimator."""
 
   def __init__(self, sketch_operator, cardinality_estimator):
--- a/src/estimators/stratified_sketch.py
+++ b/src/estimators/stratified_sketch.py
@@ -37,7 +37,8 @@ class StratifiedSketch(object):
 
     Args:
       max_freq: the maximum targeting frequency level. For example, if it is set
-        to 3, then the sketches will include frequency=1, 2, and frequency >= 3.
+        to 3, then the sketches will include frequency=1+, 1, 2, and frequency
+        >= 3.
       random_seed: This arg exists in order to conform to
         simulator.EstimatorConfig.sketch_factory.
       cardinality_sketch_factory: A cardinality sketch factory.
--- a/src/estimators/stratified_sketch.py
+++ b/src/estimators/stratified_sketch.py
@@ -14,6 +14,8 @@
 
 import copy
 import functools
+import numpy as np
+import warnings
 from wfa_cardinality_estimation_evaluation_framework.estimators.base import SketchBase
 from wfa_cardinality_estimation_evaluation_framework.estimators.base import EstimatorBase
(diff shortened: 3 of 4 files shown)


src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,242 @@
...
+  """A frequency sketch that contains cardinality sketches per frequency bucket."""
+
+  @classmethod
+  def get_sketch_factory(cls):
+
+    def f(random_seed):
+      return cls(random_seed=random_seed)
+
+    return f
+  def __init__(self, max_freq, sketch_list, random_seed=None):
```
matthewclegg, line 35: Add blank line

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,242 @@
...
+    """Construct a Stratified sketch.
+
+    Args:
+      max_freq: the maximum targeting frequency level. For example, if
+        it is set to 3, then the sketches will include frequency=1, 2, and
+        frequency >= 3.
+      random_seed: This arg exists in order to conform to
+        simulator.EstimatorConfig.sketch_factory.
+      sketch_list: List of cardinality sketches of the same sketch type.
+
```
matthewclegg, line 45: Delete blank line

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,242 @@
...
+    SketchBase.__init__(self)
+    # A dictionary that contains multiple sketches, which include:
+    # (1) sketches with frequency equal to k, where k < max_freq;
+    # (2) a sketch with frequency greater than or equal to max_freq;
+    # (3) a 1+ reach sketch.
+    self.sketches = {}
+    self.seed = random_seed
+    self.max_freq = max_freq
+
+  def init_from_exact_multi_set(self, exact_multi_set, CardinalitySketch=None):
```
matthewclegg, line 56: Would it be better to make this a class method, which then instantiates a new StratifiedSketch object?

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,242 @@
...
+    """Initialize a Stratified sketch from one ExactMultiSet
+
+    Args:
+      exact_multi_set: ExactMultiSet object to use for initialization.
+      CardinalitySketch: Class type of cardinality sketches this stratified
+        sketch will hold.
+
+    """
+    self.sketches[ONE_PLUS] = CardinalitySketch()
```
matthewclegg, line 65: Would it be a good idea to first assert that CardinalitySketch is not none?

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,242 @@
...
+    for freq in range(1, self.max_freq+1):
+        self.sketches[str(freq)] = CardinalitySketch()
+
+    for id in exact_multi_set.ids():
+      id_freq = min(exact_multi_set.frequency(id), self.max_freq)
+      self.sketches[str(id_freq)].add(id)
+      self.sketches[str(ONE_PLUS)].add(id)
+
+  def init_from_set_generator(self, set_generator, CardinalitySketch=None):
```
matthewclegg, line 74: Perhaps this could be a class method too?

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,242 @@
...
+    """Initialize a Stratified sketch from a Set Generator.
+
+    Args:
+      set_generator: SetGenerator object to draw ids from for initialization.
+      CardinalitySketch: Class type of cardinality sketches this stratified
+        sketch will hold.
+
```
matthewclegg, line 81: Delete blank line

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,242 @@
...
+    Returns:
+      A merged StratifiedSketch from the input.
+    """
+    assert isinstance(this, StratifiedSketch)
+    this.assert_compatible(that)
+    max_freq = this.max_freq
+    merged_sketch = copy.deepcopy(this)
+
+    for k in range(max_freq - 1):
```
matthewclegg, line 149: I think the code would be clearer if this were range(1, max_freq)

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,242 @@
...
+      # Calculate A(k) & B(0) = A(k) - (A(k) & B(1+))
+      merged = self.sketch_difference(
+          this.sketches[str(k + 1)], self.sketch_intersection(
```
matthewclegg, line 152: I understand why you are using ONE_PLUS as a key for the sketch consisting of all items, but I don't understand the benefit of converting the other frequencies to strings.  Why not just index on the frequency itself?

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,242 @@
...
+              this.sketches[str(k + 1)], that.sketches[ONE_PLUS]))
+
+      # Calculate A(0) & B(k) = B(k) - (B(k) & A(1+))
+      merged = self.sketch_union(
+          merged, self.sketch_difference(
+              that.sketches[str(k + 1)], self.sketch_intersection(
+                  this.sketches[ONE_PLUS], that.sketches[str(k + 1)])))
+
+      # Calculate A(i) & B(k-i)
+      for i in range(k):
```
matthewclegg, line 162: Here, it would be clearer to use range(1,k+1)

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,242 @@
...
+                that.sketches[str(max_freq)],
+                this.sketches[ONE_PLUS])))
+    merged_sketch.sketches[str(max_freq)] = merged
+
+    # Calculate Merged(1+)
+    merged_one_plus = None
+    for k in range(max_freq):
+      merged_one_plus = self.sketch_union(merged_one_plus,
+                                          merged_sketch.sketches[str(k + 1)])
+    merged_sketch.sketches[ONE_PLUS] = merged_one_plus
```
matthewclegg, line 191: Is it the case that
  merged_one_plus = self.sketch_union(this.sketches[ONE_PLUS], that.sketches[ONE_PLUS])?

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,242 @@
...
+  def difference(cls, this, that):
+    if this is None:
+      return None
+    if that is None:
+      return copy.deepcopy(this)
+    result = copy.deepcopy(this)
+    result_key_set = set(result.ids().keys())
+    that_key_set = set(that.ids().keys())
+    result._ids = {x:1 for x in result_key_set.difference(that_key_set)}
+    return result
```
matthewclegg, line 242: Why not add these methods to ExactMultiSet?  This would reduce the number of classes by one, and it would keep all of the code relating to ExactMultiSet in one place.

src/estimators/tests/stratified_sketch_test.py
```diff
@@ -0,0 +1,170 @@
...
+    universe_size = 1000
+    set_sizes = [100] * 5
+    max_freq = 3
+
+    ind_gen = set_generator.IndependentSetGenerator(
+      universe_size=universe_size, set_sizes=set_sizes,
+      random_state=np.random.RandomState(1))
+    ind_gen_clone = set_generator.IndependentSetGenerator(
+      universe_size=universe_size, set_sizes=set_sizes,
+      random_state=np.random.RandomState(1))
```
matthewclegg, line 59: I would favor using a fake SetGenerator.  It allows you to control exactly what the resulting data structure should look like.  For an example, see lines 83-100 of simulations/tests/simulator_test.py.

src/estimators/tests/stratified_sketch_test.py
```diff
@@ -0,0 +1,170 @@
...
+from absl.testing import parameterized
+
+import farmhash
+import numpy as np
+
+from wfa_cardinality_estimation_evaluation_framework.estimators import stratified_sketch
+from wfa_cardinality_estimation_evaluation_framework.estimators.exact_set import ExactMultiSet
+from wfa_cardinality_estimation_evaluation_framework.simulations import set_generator
+
+ONE_PLUS = '1+'
```
matthewclegg, line 27: It's better to refer to the definition of ONE_PLUS in stratified_sketch.py than to redefine it here.

src/estimators/tests/stratified_sketch_test.py
```diff
@@ -0,0 +1,170 @@
...
+  def test_merge_sketches(self):
+    max_freq = 3
+    this_multi_set = self.generate_multi_set([(1,2),(2,3),(3,1)])
+    that_multi_set = self.generate_multi_set([(1,1),(3,1),(4,5),(5,1)])
+    expected = {
+    ONE_PLUS : {1: 1, 2: 1, 3: 1, 4: 1, 5: 1},
+    '1' : {5: 1},
+    '2' : {3: 1},
+    '3' : {1: 1, 2: 1, 4: 1},
+    }
```
matthewclegg, line 108: Is this indented properly?

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,242 @@
...
+
+    """
+    self.sketches[ONE_PLUS] = CardinalitySketch()
+    for freq in range(1, self.max_freq+1):
+        self.sketches[str(freq)] = CardinalitySketch()
+
+    for id in exact_multi_set.ids():
+      id_freq = min(exact_multi_set.frequency(id), self.max_freq)
+      self.sketches[str(id_freq)].add(id)
+      self.sketches[str(ONE_PLUS)].add(id)
```
huangxichen1, line 72: I would prefer to use add_ids(exact_multi_set.ids()) instead of add(id). One reason is to be compatible with VoC. Every time we call voc.add(id), we need to know if we need to add this id to the voc sketch. Yet the current implementation of voc is memoryless in terms of whether this id has been added to the sketch. This comment in the following link will explain this in more detail:
https://github.com/world-federation-of-advertisers/cardinality_estimation_evaluation_framework/blob/master/src/estimators/vector_of_counts.py#L106

src/estimators/stratified_sketch.py
huangxichen1, line 56: Several questions about the CardinalitySketch.

1. Is CardinalitySketch a class? How can we make sure that the sketch instance created from the CardinalitySketch are compatible, or namely using the same random seed and sketch parameters?

On the other hand, is it possible to use a sketch_factory so as to make sure that the sketches are compatible? For example,
```
def init_from_exact_multi_set(self, exact_multi_set, cardinality_sketch_factory):
  self.sketches[ONE_PLUS] = cardinality_sketch_factory(a_random_seed)
  for freq in range(1, self.max_freq + 1):
    self.sketches[str(freq)] = cardinality_sketch_factory(a_random_seed)
```

2. For either `cardinality_sketch_factory` or `CardinalitySketch`, do we want to make it an argument in the constructor (__init__)?

3. One more thing I am not sure is about the code style. When we use a class as an argument, do we use variable_name style or VariableName style?

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,242 @@
...
+    return merged
+
+  def merge_sketches(self, this, that):
+    """Merge two StratifiedSketch.
+
+       Given 2 sketches A and B:
+       Merged(k) = (A(k) & B(0)) U (A(k-1) & B(1)) ... U (A(0) & B(k))
+         where
+         A(k) & B(0) =  A(k) - (A(k) & B(1+))
+         B(k) & A(0) =  B(k) - (B(k) & A(1+))
```
huangxichen1, line 136: The explanation is nice!

src/estimators/tests/stratified_sketch_test.py
```diff
@@ -0,0 +1,170 @@
...
+    ind_gen = set_generator.IndependentSetGenerator(
+      universe_size=universe_size, set_sizes=set_sizes,
+      random_state=np.random.RandomState(1))
+    ind_gen_clone = set_generator.IndependentSetGenerator(
+      universe_size=universe_size, set_sizes=set_sizes,
+      random_state=np.random.RandomState(1))
+    s = stratified_sketch.StratifiedSketch(sketch_list=[], max_freq=max_freq,
+          random_seed=np.random.RandomState(1))
+
+
```
huangxichen1, line 63: One empty line?

src/estimators/tests/stratified_sketch_test.py
huangxichen1, line 108: Indent by 4 spaces?

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,242 @@
...
+import copy
+import functools
+import sys
+import numpy as np
+from wfa_cardinality_estimation_evaluation_framework.estimators.base import SketchBase
+from wfa_cardinality_estimation_evaluation_framework.estimators.exact_set import ExactMultiSet
+
+ONE_PLUS = '1+'
+
+class StratifiedSketch(SketchBase):
```
huangxichen1, line 25: I am not sure if StratifiedSketch should inherit from the SketchBase. Seems like it is pretty different from the sketches we usually refer to, eg, bloom filters. I would like to hear more opinions from other reviewers.

src/estimators/tests/stratified_sketch_test.py
uakyol: definetly, thanks!

src/estimators/stratified_sketch.py
uakyol: actually, I modified it. to range(1, k) and changed the code below a little. I think this is cleaner, what do you think?

src/estimators/stratified_sketch.py
uakyol: @huangxichen1 I aggree with Matthew, that it should be the case. What do you think?

src/estimators/stratified_sketch.py
uakyol: @matthewclegg what do you think?

src/estimators/stratified_sketch.py
uakyol: @huangxichen1 we had this conversation and I think your reasoning is to keep it this way because " The union and set_difference are applied to a specific sets relationship, which can be different from a more generic situations"
 
Matthew, maybe we should keep it this way because the other sketches will need other operators which  @huangxichen1 is going to work on.

src/estimators/stratified_sketch.py
uakyol: I agree, added a add_ids function to ExactMultiSet class

uakyol: updated per our conversation offline

src/estimators/stratified_sketch.py
uakyol: yes, I agree thanks for pointing out.

src/estimators/tests/stratified_sketch_test.py
uakyol: This is so helpful, was looking for this. Thanks!

src/estimators/stratified_sketch.py
matthewclegg: I agree with Xichen.

src/estimators/stratified_sketch.py
uakyol: 1. Per our conversation offline, changed it to a sketch factory.

src/estimators/exact_set.py
```diff
@@ -53,6 +53,11 @@ def __contains__(self, x):
     """Return true if x is contained in the sketch."""
     return x in self._ids
 
+  def add_ids(self, x):
+    """Adds a list of ids to the sketch."""
+    for item_id in x:
+      self.add(item_id)
+
```
matthewclegg, line 60: Doesn't ExactMultiSet already inherit this from SketchBase?

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,261 @@
...
+
+  def __call__(self, sketches_list):
+    return self.merge_sketches(sketches_list)
+
+  def merge_sketches(self, sketches_list):
+    return functools.reduce(self.pairwise_estimator.merge_sketches,
+                            sketches_list)
+
+
+class ExactMultiSetOperation(object):
```
matthewclegg, line 223: Are there unit tests for this class?

src/estimators/exact_set.py
uakyol: you are right, sorry, fixed it.

uakyol:
Hello @my2457 I just added you as a reviewer, would love to hear your feedback.
Thanks.

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,261 @@
...
+    assert isinstance(other,
+                      StratifiedSketch), ('other is not a StratifiedSketch.')
+    assert self.seed == other.seed, ('The random seeds are not the same: '
+                                     f'{self.seed} != {other.seed}')
+    assert self.max_freq == other.max_freq, (
+        'The frequency targets are different: '
+        f'{self.max_freq} != {other.max_freq}')
+
+
+class PairwiseEstimator(object):
```
my2457, line 124: A little bit confused with this name. Will this be used as estimator for this frequency-based bloom filter? We probably need to inherit `EstimatorBase` and need to return a list in __call__?

src/estimators/stratified_sketch.py
matthewclegg, line 223: Should this be called ExactMultiSetOperation?  Perhaps a better name would be ExactSetOperation, since the frequencies are being set to one?

uakyol: yes, I was thinking about that too. I was just not sure since we don't have an ExactSet class. 
Changed.

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,261 @@
...
+      return cls(random_seed=random_seed)
+
+    return f
+
+  def __init__(self, max_freq, cardinality_sketch_factory, random_seed):
+    """Construct a Stratified sketch.
+
+    Args:
+      max_freq: the maximum targeting frequency level. For example, if it is set
+        to 3, then the sketches will include frequency=1, 2, and frequency >= 3.
```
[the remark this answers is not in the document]
shengma-1: The alternative is to do 1, 2, 3+. Not sure which structure is better as the local sketch has noise.

src/estimators/stratified_sketch.py
uakyol: Thanks for pointing this out. Yes, this should be the case. I was thinking of adding the estimation in a following PR but this seems like a good place to do it. I added the estimation as well.

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,294 @@
...
+from wfa_cardinality_estimation_evaluation_framework.estimators.exact_set import ExactMultiSet
+
+ONE_PLUS = '1+'
+
+
+class StratifiedSketch(object):
+  """A frequency sketch that contains cardinality sketches per frequency bucket."""
+
+  @classmethod
+  def get_sketch_factory(cls):
```
kungfucraig, line 28: This method is untested, and I believe it is also broken.

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,294 @@
...
+import copy
+import functools
+from wfa_cardinality_estimation_evaluation_framework.estimators.base import SketchBase
+from wfa_cardinality_estimation_evaluation_framework.estimators.base import EstimatorBase
+from wfa_cardinality_estimation_evaluation_framework.estimators.exact_set import ExactMultiSet
+
+ONE_PLUS = '1+'
+
+
+class StratifiedSketch(object):
```
kungfucraig, line 24: How do we expect this class to integrate with the Simulator?

In its current form I think it may be incompatible.

Can you update https://github.com/world-federation-of-advertisers/cardinality_estimation_evaluation_framework/blob/master/examples/basic_comparison.py to include StratifiedSketch?

@shengma-1 @matthewclegg 

Also, it should inherit SketchBase.

matthewclegg: Updating basic_comparison.py sounds like a nice idea.  For example, we could add a StratifiedSketch of ExactMultiSets as a verification.

Not sure that I agree with the suggestion of inheriting from SketchBase.  SketchBase includes the add(id) method, and there is not an obvious implementation of add(id) for StratifiedSketches.

matthewclegg: Also, FYI, I think there are at least 2 more PR's in the pipeline that relate to the StratifiedSketch, so this is not the final word.  Everything will integrate in the end.

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,294 @@
...
+     stratified_sketch: a StratifiedSketch object.
+
+    Returns:
+      A dictionary: the key is the frequency and the value is the corresponding
+      cardinality.
+    """
+    result = {}
+    for freq, sketch in stratified_sketch.sketches.items():
+      result[freq] = self.cardinality_estimator([sketch])
+    return result
```
matthewclegg, line 237: This should return a cumulative histogram.  Also, self.cardinality_estimator([sketch]) itself returns a histogram, which makes me suspect that this is not being tested?

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,294 @@
...
+
+
+class SequentialEstimator(EstimatorBase):
+  """Sequential frequency estimator."""
+
+  def __init__(self, sketch_operator, cardinality_estimator):
+    self.pairwise_estimator = PairwiseEstimator(sketch_operator,
+                                                cardinality_estimator)
+
+  def __call__(self, sketches_list):
```
matthewclegg, line 247: This should return the cumulative histogram, which is a list.  However, self.pairwise_estimator.estimate_cardinality returns a dictionary.  See the class definition of EstimatorBase in src/estimators/base.py.

uakyol: Thanks for pointing this out, I was returning a dict, it is now returning a cumulative histogram.

src/estimators/stratified_sketch.py
uakyol: It was being tested, was assuming a dict output though. Now we are outputting a cumulative histogram.

src/estimators/stratified_sketch.py
shengma-1: The VoC/ADBF frequency dedupe will this class. See a new PR #44.

uakyol: This is now integrated with the simulator.

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,348 @@
...
+        methods for two sketches.
+      cardinality_estimator: a cardinality estimator for estimating the
+        cardinality of a sketch.
+    """
+    self.cardinality_estimator = cardinality_estimator
+    self.sketch_union = sketch_operator.union
+    self.sketch_difference = sketch_operator.difference
+    self.sketch_intersection = sketch_operator.intersection
+
+  def __call__(self, this, that):
```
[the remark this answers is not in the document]
uakyol: ah, thanks for this, I had added and then removed accidentally it seems. Also added a test for this.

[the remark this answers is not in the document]
matthewclegg: Perhaps having them in the merge() function is better.

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,355 @@
...
+        max_freq=max_freq,
+        underlying_set=exact_multi_set,
+        cardinality_sketch_factory=cardinality_sketch_factory,
+        random_seed=random_seed)
+    stratified_sketch.create_sketches()
+
+    return stratified_sketch
+
+  @classmethod
+  def init_from_set_generator(cls, max_freq, set_generator,
```
kungfucraig, line 121: Do we expect this method to get called? It seems to me maybe not.

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,355 @@
...
+
+    Args:
+      other: the other StratifiedSketch for comparison.
+
+    Raises:
+      AssertionError: if the other sketches are not StratifiedSketch, or if
+      their random_seed are different, or if the frequency targets are
+      different.
+    """
+    assert isinstance(other,
```
kungfucraig, line 150: Should they have the same cardinality sketch factory? What about underlying set type?

examples/basic_comparison.py
```diff
@@ -101,6 +106,16 @@ def main(argv):
       sketch_factory=ExactMultiSet.get_sketch_factory(),
       estimator=LosslessEstimator())
 
+  estimator_config_stratified = SketchEstimatorConfig(
+      name='stratified',
```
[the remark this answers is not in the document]
uakyol: Done

src/estimators/stratified_sketch.py
uakyol: Oh definitely, it is the main use case Matthew envisioned.

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,355 @@
...
+          underlying_set=underlying_set,
+          cardinality_sketch_factory=cardinality_sketch_factory)
+
+    return f
+
+  def __init__(self,
+               max_freq,
+               cardinality_sketch_factory,
+               random_seed,
+               underlying_set=None):
```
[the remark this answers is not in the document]
uakyol: Done

src/estimators/stratified_sketch.py
```diff
@@ -0,0 +1,355 @@
...
+
+
+class StratifiedSketch(SketchBase):
+  """A frequency sketch that contains cardinality sketches per frequency bucket."""
+
+  @classmethod
+  def get_sketch_factory(cls,
+                         max_freq,
+                         cardinality_sketch_factory,
+                         underlying_set=None):
```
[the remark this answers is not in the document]
uakyol: Done

examples/basic_comparison.py
```diff
@@ -120,10 +136,11 @@ def main(argv):
       'exact_set': estimator_config_exact,
       'hll++': estimator_config_hll,
       'vector_of_counts': estimator_config_voc,
+      'stratified_sketch': estimator_config_stratified,
```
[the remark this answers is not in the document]
uakyol: done

src/estimators/stratified_sketch.py
uakyol: asserted the types

PR #40, opened 2020-08-12
Packaged run_evaluation.py as the command line script 'wfa-run-evaluation'

pfhgetty:
This pull request enables the functionality of run_evaluation.py to be run from anywhere as:
`wfa-run-evaluation <flags>` 
as long as the cardinality_estimation_evaluation_framework is installed. 

This allows end users to install this repository as a pip package and run the evaluations without having to access src/evaluations/run_evaluation.py or create their own script which imports wfa_cardinality_estimation_evaluation.evaluations.run_evaluation. 

Modified docstring in run_evaluation.py to reflect this change. Also added `--num_runs=1` to the example command to run the evaluations.


setup.py
```diff
@@ -27,4 +27,11 @@
         'wfa_cardinality_estimation_evaluation_framework.evaluations',
         'wfa_cardinality_estimation_evaluation_framework.evaluations.data',
         'wfa_cardinality_estimation_evaluation_framework.simulations',
-    ])
+    ],
+    entry_points={
+        'console_scripts': [
+            'run-evaluation=wfa_cardinality_estimation_evaluation_framework'
```
jgoodknight, line 33: I wonder if the name of the global script should be a bit more specific like wfa-run-evaluation so it's even more clear the console script comes from this repo, and tab completion can work off of wfa instead of run?

pfhgetty: Resolved in 623deb4

src/evaluations/run_evaluation.py
```diff
@@ -14,12 +14,13 @@
 r"""A commandline tool to run evaluations of cardinality estimators.
 
 To run this evaluation, in command line, type:
-python run_evaluation.py \
---evaluation_out_dir="evaluation_output" \
---analysis_out_dir="analysis_output" \
---evaluation_config="smoke_test" \
---sketch_estimator_configs="vector_of_counts-4096-ln3-sequential" \
---evaluation_run_name="simple_run"
+  wfa-run-evaluation \
```
kungfucraig, line 17: The other way will still be used by those who develop.

I think maybe you want this documentation in main.py?

With perhaps a note here to look at main.py for those who have the environment installed?

pfhgetty: I actually intend for this to be used by developers, too. Is there a reason that devs should use the other way? 

Encouraging devs to use this way will ensure that the version of run_evaluation.py they are using always matches up with the version of the package they have installed. For example, when switching branches, running `python run_evaluation.py` without first running `python setup.py install` will run the run_evaluation.py script of the current branch while running the code of the previous branch which could cause nontrivial, hard-to-debug behaviors. Running `wfa-run-evaluation`, on the other hand, will always use the run_evaluation.py that is installed in the current environment, ensuring consistent behavior while developing.

kungfucraig: Makes sense. Thanks for the explanation.

PR #41, opened 2020-08-17
Defines the IndependentSet estimator.

matthewclegg:
This PR defines an estimator that can be used for estimating the cardinality of a union under the independence assumption.  In a later PR, I will provide an example config that makes use of this estimator.


src/estimators/independent_set_estimator.py
```diff
@@ -0,0 +1,93 @@
...
+
+    Args:
+      single_sketch_estimator:  An object derived from EstimatorBase for
+        providing cardinality estimates of individual sketches.
+      universe_size: Size of the universe.
+    """
+    self.single_sketch_estimator = single_sketch_estimator
+    self.universe_size = universe_size
+
+  def __call__(self, sketch_list):
```
shengma-1, line 49: sketch_list requires frequency sketch representation? Need to sync on this structure. The stratefied sketch uses per freq sketch.

matthewclegg: I don't think there is a conflict with StratifiedSketch here.

src/estimators/independent_set_estimator.py
```diff
@@ -0,0 +1,93 @@
...
+      for i in range(len(b_hist)):
+        c_hist[i] += b_hist[i]
+      for i in range(len(a_hist)):
+        for j in range(len(b_hist)):
+          overlap = a_hist[i] * b_hist[j] / self.universe_size
+          if overlap:
+            c_hist[i] -= overlap
+            c_hist[j] -= overlap
+            if len(c_hist) < i+j+2:
+              c_hist += [0] * (i + j + 2 - len(c_hist))
```
shengma-1, line 78: will this too slow to dynamic resize? Alternative, initialize c_hist before line71.

src/estimators/independent_set_estimator.py
```diff
@@ -0,0 +1,93 @@
...
+            c_hist[i+j+1] += overlap
+
+      a_hist = c_hist
+
+      if sum(a_hist) > self.universe_size:
```
kungfucraig, line 83: Probably use "assert" here so that you'd raise an AssertionError.

ValueError is for signaling bad input arguments. From Python docs: "Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError."

src/estimators/independent_set_estimator.py
matthewclegg: I moved the code, although for the record, the asymptotic runtime is unchanged.

src/estimators/independent_set_estimator.py
matthewclegg: Done.

src/estimators/independent_set_estimator.py
```diff
@@ -0,0 +1,112 @@
...
+      a_hist = c_hist
+
+      assert sum(a_hist) <= self.universe_size, (
+        "Constraint violation: sketch is larger than universe")
+
+    # At this point, a_hist is the histogram of exact frequencies,
+    # e.g., a_hist[i] is the number of items having frequency exactly i+1.
+    # Now convert it into a histogram of cumulative frequencies.
+
+    a_hist = [int(f) for f in a_hist]
```
my2457, line 104: Do we have to round it here or maybe let the caller round it similar to the estimators previously written?

src/estimators/independent_set_estimator.py
```diff
@@ -0,0 +1,112 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Estimator for sketches when using the independence assumption.
+
+Under the independence assumption, the cardinality of the union of two
+sketches representing sets A and B is equal to
+   
+  |A| + |B| - |A| * |B| / N,
+
+where N is the size of the universe from which A and B are drawn.
```
AaronRanAn, line 21: @matthewclegg, @shengma-1 can you share any reference literature for this estimator? The logic makes sense to me and it seems fine as a baseline for our evaluation, but I am just not familiar with the theory behind this, so would love to learn more.

shengma-1: The independence model assumes that ids in setA and ids in setB are independence draw from a universe (size=N).
Pr(ids in A and B) = Pr(ids in A) + Pr(ids in B) - Pr(Ids in A) * Pr(ids in B).
This paper: https://pdfs.semanticscholar.org/b0b1/34bf4f5d59c2c4f6e67b61f64dc1a1474646.pdf?_ga=2.21486968.293676905.1597883007-230818821.1551825210 shows that cross media reach is well estimated by the independence model.

shengma-1: Sorry, the equation was wrong. It should be:
Pr(ids in A and B) = Pr(Ids in A) * Pr(ids in B).
Pr(ids in A or B) = Pr(ids in A) + Pr(ids in B) - Pr(Ids in A) * Pr(ids in B).

AaronRanAn: Thank you Sheng! The paper is really clear.

src/estimators/independent_set_estimator.py
matthewclegg: I have removed the rounding.

PR #2, opened 2020-04-14
Google's first submit sharing what we've developed.

jgoodknight:
Google's first submit sharing what we've developed.

PR #42, opened 2020-08-18
Frequency evaluation framework set generator #2: heterogeneous user frequency

matthewclegg:


src/simulations/set_generator.py
```diff
@@ -626,3 +626,96 @@ def __iter__(self):
...
+
+  The frequency distribution is defined as a shifted-Poisson distribution.
+  I.e., freq ~ Poission(freq_rate) + 1.
+  """
+
+  @classmethod
+  def get_generator_factory_with_num_and_size(cls, universe_size, num_sets,
+                                              set_size, gamma_params_list,
+                                              freq_cap):
+
```
kungfucraig, line 652: Can you add a check for  whether the expected set_size_list == gamma_params_list here? Otherwise this error check and subsequent failure are significantly delayed. 

The stack trace you get from the error in __init__ will not easily lead you to you problem. The stack trace from here will.

src/simulations/set_generator.py
```diff
@@ -626,3 +626,96 @@ def __iter__(self):
...
+      return cls(universe_size, list(_SetSizeGenerator(num_sets, set_size)),
+                 gamma_params_list, random_state, freq_cap)
+
+    return f
+
+  @classmethod
+  def get_generator_factory_with_set_size_list(cls, universe_size,
+                                               set_size_list, gamma_params_list,
+                                               freq_cap):
+
```
kungfucraig, line 663: Ideally same error check here too.

src/simulations/set_generator.py
```diff
@@ -626,3 +626,96 @@ def __iter__(self):
...
+    assert all([gamma_params[1] > 0 for gamma_params in gamma_params_list]), (
+        'Gamma rate parameters must be positive.')
+    assert freq_cap is None or freq_cap > 0, (
+        'freq_cap should be None or positive.')
+    self.universe_size = universe_size
+    self.set_sizes = set_sizes
+    self.gamma_params_list = gamma_params_list
+    self.freq_cap = freq_cap
+    self.random_state = random_state
+    self.choice = choice_fast
```
kungfucraig, line 704: Previously "self.choice" existed so we could fake the choice algorithm in test code. Since you are not faking it I think it's reasonable to just call "choice_fast" in __iter__.

src/simulations/set_generator.py
```diff
@@ -626,3 +626,96 @@ def __iter__(self):
...
+
+    return f
+
+  @classmethod
+  def get_generator_factory_with_set_size_list(cls, universe_size,
+                                               set_size_list, gamma_params_list,
+                                               freq_cap):
+
+    def f(random_state):
+      return cls(universe_size, set_size_list, gamma_params_list, random_state,
```
kungfucraig, line 665: We really ought not to suffix these args with "_list" as they need not be lists, but just any iterable. This comment applies to pretty much everywhere we are using the "_list" suffix.

src/simulations/set_generator.py
```diff
@@ -626,3 +626,96 @@ def __iter__(self):
...
+  def __init__(self, universe_size, set_sizes, gamma_params_list, random_state,
+               freq_cap=None):
+    """Create a heterogeneous multiset generator.
+
+    Args:
+      universe_size: An integer value that specifies the size of the whole
+        universe from which the ids will be sampled.
+      set_sizes: A list containing the size of each set, aka, the number of
+        reached IDs.
+      gamma_params_list: A list of the same size as set_sizes, specifying the
```
kungfucraig, line 679: The code below indicates this is a list of two-tuples specifying shape and scale params.

src/simulations/set_generator.py
```diff
@@ -626,3 +626,96 @@ def __iter__(self):
       self.random_state.shuffle(multiset_ids)
       yield multiset_ids
     return self
+
+class HeterogeneousMultiSetGenerator(SetGeneratorBase):
+  """Heterogeneous multiset generator.
+
+  This generator returns the multisets as described in section
+  Frequency scenario 1: Heterogeneous user frequency of this doc:
```
kungfucraig, line 634: scenario 2 :)

src/simulations/set_generator.py
matthewclegg: The convention of using the parameter name 'set_size_list' to represent an iterable of set sizes seems to be present in many places throughout the code, so it's probably not worth the effort to change that now.  However, I have shorted gamma_params_list to gamma_params.

src/simulations/set_generator.py
matthewclegg: Done

src/simulations/set_generator.py
matthewclegg: Done.

src/simulations/set_generator.py
matthewclegg: Done.

src/simulations/set_generator.py
matthewclegg: Thanks for pointing this out.

src/simulations/set_generator.py
matthewclegg: Done.

src/simulations/set_generator.py
```diff
@@ -626,3 +626,102 @@ def __iter__(self):
...
+
+    self.universe_size = universe_size
+    self.freq_cap = freq_cap
+    self.random_state = random_state
+
+  def __iter__(self):
+    for set_size, gamma_params in zip(self.set_sizes, self.gamma_params):
+      set_ids = choice_fast(self.universe_size, set_size, self.random_state)
+      rate_parameters = self.random_state.gamma(shape=gamma_params[0],
+                                                scale=gamma_params[1],
```
my2457, line 716: `scale` or `rate`? I see it's indeed `rate` from comment `shape and rate parameters of the gamma distributions.`?

matthewclegg: Thanks for catching that.

PR #49, opened 2020-08-24
Added reach estimator based on FirstMomementEstimator that add global noise

kungfucraig:
Extended FirstMomentEstimator, such that when a noiser is provider, noise is added to the number of active buckets before performing the cardinality estimation. This support the global DP scenario implemented by our MPC protocol.


src/estimators/bloom_filters.py
```diff
@@ -419,6 +421,53 @@ def __call__(self, sketch_list):
...
+    union = sketch_list[0]
+    for cur_sketch in sketch_list[1:]:
+      union.assert_compatible(cur_sketch)
+      union.sketch = 1 - (1 - union.sketch) * (1 - cur_sketch.sketch)
+
+    noiser = GeometricEstimateNoiser(
+        self.epsilon, random_state=self.random_state)
+    noise = noiser(0.)
+
+    a = union.decay_rate
```
kungfucraig, line 450: This we agree to this approach I'll do some refactoring to pull out the code that this shares with the FirstMomentEstimator, or make some of the FirstMomentEstimator's methods public.

src/estimators/bloom_filters.py
```diff
@@ -419,6 +421,53 @@ def __call__(self, sketch_list):
         union, self._weights)]
 
 
+class FirstMomentGlobalNoiseEstimator(EstimatorBase):
+  """This estimator unions the input sketches, generate geometric noise,
+     appends that many fake buckets to the sketch union and calls the
+     FirstMomentEstimator.
```
shengma-1, line 427: This only supports exp-adbf, right? Do you want to make this more general for all adbf families?

An alternative is to extend FirstMomentNoiseEstimator to take a global_estimator_noiser, which, if defined, will be added to #ones inside of FirstMomentNoiseEstimator.

kungfucraig: Yeah, I think that's a reasonable way to refactor. Will send an update.

huangxichen1: Another approach is to change the AnyDistributionBloomFilter, i.e., add an optional global noiser argument to the constructor, and create a property method for the number of one bits (the `x` in the exp and log bloom filter estimator, and the `n` in the geo bloom filter). In this way, we don't need to modify all the estimator methods. Please let me know if this makes sense.

Thanks Craig!

kungfucraig: PTAL. Refactored FirstMomentEstimator based on conversation yesterday.

src/estimators/bloom_filters.py
```diff
@@ -290,12 +292,13 @@ class FirstMomentEstimator(EstimatorBase):
   METHOD_EXP = "exp"
   METHOD_ANY = "any"
 
-  def __init__(self, method, denoiser=None, weights=None):
+  def __init__(self, method, denoiser=None, noiser=None, weights=None):
```
matthewclegg, line 295: I suppose if you set the default value for noiser=lambda x:x, you could save a few lines of code.

kungfucraig: Setting the default value of the argument makes my error check unworkable, however I did use the lambda trick in the body of __init__, which did save a few lines of repeated code in other functions. Thanks!

kungfucraig:
@AaronRanAn @shengma-1 

I need to get this submitted today, so given other reviews I'm going to submit now. Please feel to review later if you'd like.

PR #53, opened 2020-08-27
Fix the bug in evaluation_configs module for the interoperability test.

huangxichen1:


kungfucraig:
Can you leave a description of the bug in the PR description?

PR #55, opened 2020-08-27
Add local and global epsilon to ADBF.

huangxichen1:


src/evaluations/data/evaluation_configs.py
```diff
@@ -498,81 +519,236 @@ def get_evaluation_config(config_name):
...
+  else:
+    sketch_noiser, sketch_denoiser = None, None
+
+  if estimate_epsilon:
+    estimate_noiser = estimator_noisers.GeometricEstimateNoiser(
+        epsilon=estimate_epsilon)
+  else:
+    estimate_noiser = None
+
+  length = int(length)
```
my2457, line 608: why do we need this?

src/evaluations/report_generator.py
```diff
@@ -225,7 +227,8 @@ def _get_boxplot_html(row):
           row[evaluation_configs.SKETCH] + ', '
           + row[evaluation_configs.ESTIMATOR]  + '<br>'
           + 'sketch_config: ' + row[evaluation_configs.SKETCH_CONFIG] + '<br>'
-          + 'epsilon: ' + row[evaluation_configs.EPSILON])
+          + 'sketch_epsilon: ' + row[SKETCH_EPSILON] + '<br>'
+          + 'estimate_epsilon: ' + row[ESTIMATE_EPSILON])
```
my2457, line 231:
```suggestion
          + f'{SKETCH_EPSILON}: ' + row[SKETCH_EPSILON] + '<br>'
          + f'{ESTIMATE_EPSILON}: ' + row[ESTIMATE_EPSILON])
```

src/evaluations/data/evaluation_configs.py
huangxichen1: To make sure that the length of the BF sketch is integer value. Otherwise, there would be error when generating the sketches. I refactored it to the ADBF_LENGTH_LIST global variables and removed this line, so that there is no need to worry about it in this function.

PR #56, opened 2020-08-27
Add conditional independent estimator config using reach estimator to the evaluation.

huangxichen1:
Add reach estimator and config to the evaluation.


src/evaluations/data/evaluation_configs.py
```diff
@@ -568,6 +569,47 @@ def construct_sketch_estimator_config_name(sketch_name, sketch_config,
 
 
 # Document the estimators.
+def _independent_set_estimator(sketch_epsilon=None, estimate_epsilon=None):
```
kungfucraig, line 572: Should there be a test for this?

huangxichen1: Added a unit test to check if this works. PTAL. Thanks!

PR #57, opened 2020-08-27
Start counting the number of estimable set from at least two sets.

huangxichen1:
When the sketch noise is large, it is possible that the estimator cannot estimate the cardinality of ONE set under the 95/5 criterion. Yet it is still possible that it can have a good performance when estimating the union of two or more sets under 95/5.


huangxichen1:
> Can you please leave a comment on the entire PR that describes why this was necessary?

Added in the description. Thanks!

PR #64, opened 2020-09-08
ExponentialSameKeyAggregator sketch

jiayu-google:
Implement the ExponentialSameKeyAggregator sketch for global frequency estimation.

This PR implements the sketch only. Will implement the estimator later.

### DIFF
--- a/src/estimators/any_sketch.py
+++ b/src/estimators/any_sketch.py
@@ -32,22 +32,6 @@ class ValueFunction(object):
   def __eq__(self, other):
     return isinstance(other, self.__class__)
 
-  METHOD_ADD_ONE = 'add_one'
-  METHOD_ADD_POSITIVE_INTEGER_KEY = 'add_positive_key'
-
-  @classmethod
-  def get_value_from_id(cls, x, method):
-    """Convert an ID to the value to be added into the sketch."""
-    assert method in (ValueFunction.METHOD_ADD_ONE,
-                      ValueFunction.METHOD_ADD_POSITIVE_INTEGER_KEY)
-    if method == ValueFunction.METHOD_ADD_ONE:
-      return 1
-    if method == ValueFunction.METHOD_ADD_POSITIVE_INTEGER_KEY:
-      assert isinstance(x, int) and (x >= 0) and (x < 2 ** 31), (
-          'Now for conveninence, we only support'
-          'adding non-negative np.int32 ids.')
-      return x + 1
-
 
 class SumFunction(ValueFunction):
 
--- a/src/estimators/any_sketch.py
+++ b/src/estimators/any_sketch.py
@@ -47,6 +47,40 @@ class BitwiseOrFunction(ValueFunction):
     return x | y
 
 
+class UniqueKeyFunction(ValueFunction):
+  """ValueFunction to track the state of unique key of a register."""
+
+  FLAG_EMPTY_REGISTER = -1
+  FLAG_COLLIDED_REGISTER = -2
+
+  def __call__(self, x, y):
+    """ValueFunction to track the state of unique key of a register.
+
+    Args:
+      x: A state of unique key. It can be either a real key (hashed ID)
+        indicating the unique key in the register, or FLAG_EMPTY_REGISTER
+        indicating that the register is empty, or FLAG_COLLIDED_REGISTER
+        indicating that the register already has collision.
+      y: Another state of unique key.
+
+    Returns:
+      A state of unique key that merges x and y.
+    """
+    if x == UniqueKeyFunction.FLAG_EMPTY_REGISTER:
+      return y
+    if y == UniqueKeyFunction.FLAG_EMPTY_REGISTER:
+      return x
+    if (x == UniqueKeyFunction.FLAG_COLLIDED_REGISTER) or (
+        y == UniqueKeyFunction.FLAG_COLLIDED_REGISTER):
+      return x
+    # When x & y are neither FLAG_EMPTY_REGISTER or FLAG_COLLIDED_REGISTER,
+    # i.e., they are real keys,
+    # it suffices to check if they are the same keys.
+    if x == y:
+      return x
+    return UniqueKeyFunction.FLAG_COLLIDED_REGISTER
+
+
 class Distribution(object):
   """A base class for distributions to document interface."""
 
@@ -272,6 +306,8 @@ class AnySketch(SketchBase):
     self.sketch = np.zeros(
         tuple(len(i.distribution) for i in config.index_specs),
         dtype=np.int32)
+    if config.value_functions[0] == UniqueKeyFunction():
+      self.sketch += UniqueKeyFunction.FLAG_EMPTY_REGISTER
     # We create config.num_hashes * #indexes hashes. Idealy we would
     # only need one hash per index dimension, but multiple makes the
     # implementation easier. There is probably a better way that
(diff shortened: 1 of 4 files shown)


src/estimators/any_sketch.py
```diff
@@ -47,6 +47,40 @@ def __call__(self, x, y):
...
+      y: Another state of unique key.
+
+    Returns:
+      A state of unique key that merges x and y.
+    """
+    if x == UniqueKeyFunction.FLAG_EMPTY_REGISTER:
+      return y
+    if y == UniqueKeyFunction.FLAG_EMPTY_REGISTER:
+      return x
+    if (x == UniqueKeyFunction.FLAG_COLLIDED_REGISTER) or (
```
kungfucraig, line 73: You don't need the parens around each of the terms of the "or".

src/estimators/any_sketch.py
```diff
@@ -47,6 +47,40 @@ def __call__(self, x, y):
     return x | y
 
 
+class UniqueKeyFunction(ValueFunction):
+  """ValueFunction to track the state of unique key of a register."""
+
+  FLAG_EMPTY_REGISTER = -1
```
kungfucraig, line 53: Can this be zero? It simplifies a lot of code below if it can.

src/estimators/any_sketch.py
```diff
@@ -272,6 +306,8 @@ def __init__(self, config, random_seed=None, hash_function=HashFunction):
     self.sketch = np.zeros(
         tuple(len(i.distribution) for i in config.index_specs),
         dtype=np.int32)
+    if config.value_functions[0] == UniqueKeyFunction():
```
kungfucraig, line 309: This has the potential to cause problems. If we can make the FLAG_EMPTY_REGISTER == 0, then we can delete this.

src/estimators/any_sketch.py
```diff
@@ -317,8 +353,18 @@ def add(self, x):
     indexes = self.get_indexes(x)
     # Move this to the loop when we support more than one value function.
     value_function = self.config.value_functions[0]
+
+    if value_function == UniqueKeyFunction():
+      assert self.config.num_hashes == 1, (
+          'Same Key Aggregator only supports one hash function.')
+      assert isinstance(x, int) and (x >= 0) and (x < 2 ** 31), (
```
kungfucraig, line 360: Our hash function uses an int64. We may just want to udpate the dtype across the board for this.

src/estimators/any_sketch.py
```diff
@@ -317,8 +353,18 @@ def add(self, x):
     indexes = self.get_indexes(x)
     # Move this to the loop when we support more than one value function.
     value_function = self.config.value_functions[0]
+
+    if value_function == UniqueKeyFunction():
+      assert self.config.num_hashes == 1, (
```
kungfucraig, line 358: Why have this limitation?

src/estimators/any_sketch.py
```diff
@@ -317,8 +353,18 @@ def add(self, x):
...
+          'Same Key Aggregator only supports one hash function.')
+      assert isinstance(x, int) and (x >= 0) and (x < 2 ** 31), (
+          'Now for conveninence, Same Key Aggregator only supports'
+          'adding integer np.int32 ids.')
+      value_to_add = x
+    else:
+      value_to_add = 1
     for index in indexes:
-      self.sketch[index] = value_function(self.sketch[index], 1)
+      self.sketch[index] = value_function(self.sketch[index], value_to_add)
```
kungfucraig, line 367: I think it would be better to add a method on value function that takes the input value, "x" and returns the value to add. The existing ones can just return 1. Maybe call that method "get_value_from_fingerprint"?

kungfucraig (changes requested):

src/estimators/same_key_aggregator.py
```diff
@@ -0,0 +1,56 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+"""Contains the class AnySketch and the objects required to configure it."""
+
+from wfa_cardinality_estimation_evaluation_framework.estimators import any_sketch
+
+
+class ExponentialSameKeyAggregator():
```
huangxichen1, line 19: class ExponentialSameKeyAggregator(SketchBase) or class ExponentialSameKeyAggregator?

src/estimators/any_sketch.py
jiayu-google: Good point. I was setting it as -1 because our ID (from set_generator) can be 0. 

In this version, I just set FLAG_EMPTY_REGISTER as 0, so we don't need to change any_sketch.

src/estimators/any_sketch.py
jiayu-google: Thanks, removed.

src/estimators/any_sketch.py
jiayu-google: Yes, See comment above.

src/estimators/any_sketch.py
jiayu-google: Removed this in the updated version.

Was adding this limitation because I could not think of an estimator for SKA with multiple hashes...

src/estimators/any_sketch.py
jiayu-google: I'm keep int32 here because I add raw IDs as the keys (which are smaller than 2^31 ~ 2 billion).

If you prefer adding the hashed ID rather than raw ID, I will make the change!

src/estimators/any_sketch.py
jiayu-google: good suggestion, added!

src/estimators/same_key_aggregator.py
jiayu-google: Thanks! Added (SketchBase)

jiayu-google:
Re-organized the codes so that
- the base class AnySketch is untouched;
- all the changes occur in any_sketch.UniqueKeyFunction and same_key_aggregator.py

Thanks to Xichen's kind help and to Craig's comments.

jgoodknight (approved):

huangxichen1 (approved):

src/estimators/any_sketch.py
```diff
@@ -47,6 +47,54 @@ def __call__(self, x, y):
...
+    if y == UniqueKeyFunction.FLAG_EMPTY_REGISTER:
+      return x
+    if x == UniqueKeyFunction.FLAG_COLLIDED_REGISTER or y == UniqueKeyFunction.FLAG_COLLIDED_REGISTER:
+      return UniqueKeyFunction.FLAG_COLLIDED_REGISTER
+    # When x & y are neither FLAG_EMPTY_REGISTER or FLAG_COLLIDED_REGISTER,
+    # i.e., they are real keys,
+    # it suffices to check if they are the same keys.
+    if x == y:
+      return x
+    return UniqueKeyFunction.FLAG_COLLIDED_REGISTER
```
my2457, line 80: Sounds the logic can be simplified? Considering it's `FLAG_EMPTY_REGISTER` only if both x and y == FLAG_EMPTY_REGISTER
```suggestion
    if x == y:
      return x
    return UniqueKeyFunction.FLAG_COLLIDED_REGISTER
```

my2457 (approved):

src/estimators/any_sketch.py
jiayu-google: Thanks! Please see the revision with simplified logic and additional comments.

src/estimators/any_sketch.py
```diff
@@ -47,6 +47,54 @@ def __call__(self, x, y):
...
+      y: Another state of unique key.
+
+    Returns:
+      A state of unique key that merges x and y.
+    """
+    if x == UniqueKeyFunction.FLAG_EMPTY_REGISTER:
+      return y
+    if y == UniqueKeyFunction.FLAG_EMPTY_REGISTER:
+      return x
+    if x == UniqueKeyFunction.FLAG_COLLIDED_REGISTER or y == UniqueKeyFunction.FLAG_COLLIDED_REGISTER:
```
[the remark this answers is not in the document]
jiayu-google: Sure, made it by calling UniqueKeyFunction.FLAG_COLLIDED_REGISTER as a temporary variable "collision"...

jiayu-google:
Thanks for your comments!

PR #66, opened 2020-09-16
Increase universe size.

huangxichen1:
This PR increases the universe size from 100k to 1m for the complete evaluation.

The rationale is that the sketch tested are designed for the US internet population which is far larger than 100k. The small universe, and hence the small set size, will lead to a very sparse sketch, which causes us trouble in the local differential private scenario. That is, the signal (IDs added to the sketch) is relatively small compared to the noise. The number of estimable sets of ADBF under the small universe will be close to 0, which is misleading.


jiayu-google (approved):

my2457 (approved):
Should we also update the doc? It's kind of deviating from the doc specs for different scenarios

AaronRanAn (approved):
LGTM

kungfucraig (approved):

PR #68, opened 2020-09-17
Add stratified sketch (voc) configs.

huangxichen1:


src/evaluations/evaluator.py
```diff
@@ -321,8 +321,8 @@ def save_estimator(self, sketch_estimator_config):
     estimator = sketch_estimator_config.sketch_factory(0)
     sketch_estimator_config_file = os.path.join(
         estimator_dir, ESTIMATOR_CONFIG_FILE)
-    with open(sketch_estimator_config_file, 'wb') as f:
-      pickle.dump(estimator, f)
+    # with open(sketch_estimator_config_file, 'wb') as f:
+    #   pickle.dump(estimator, f)
```
my2457, line 325: why comment this?

jiayu-google (approved):

src/evaluations/evaluator.py
huangxichen1: The pickle module cannot pickle classes that are not defined at the top level of a module. As the sketch estimator config we defined here wraps the voc sketch_factory method which returns a class and is not at the top level, the pickle will fail. The error message reads
```
AttributeError: Can't pickle local object 'VectorOfCounts.get_sketch_factory.**<locals>**.f'
```

I will remove this change in this PR though. Instead, I will create another PR to remove the pickle module since it is not used by the analyzer and report_generator anyway.

FYI: https://docs.python.org/3/library/pickle.html#what-can-be-pickled-and-unpickled

my2457 (approved):

AaronRanAn (approved):
LGTM

uakyol (approved):

src/evaluations/data/evaluation_configs.py
```diff
@@ -966,19 +968,86 @@ def _generate_cardinality_estimator_configs():
...
+      name=construct_sketch_estimator_config_name(
+          sketch_name='stratified_sketch_vector_of_counts',
+          sketch_config='4096',
+          estimator_name=f'sequential_{clip_str}',
+          sketch_epsilon=sketch_epsilon,
+          max_frequency=str(max_frequency)),
+      sketch_factory=stratified_sketch.StratifiedSketch.get_sketch_factory(
+          max_freq=max_frequency,
+          cardinality_sketch_factory=(
+              vector_of_counts.VectorOfCounts.get_sketch_factory(4096)
```
kungfucraig, line 1004: I think 4096 should be an input to the config function.

kungfucraig (approved):
Tests?

src/evaluations/data/evaluation_configs.py
huangxichen1: Added tests. Also added length as the argument.

PR #74, opened 2020-09-22
Fix bug for voc sketch operator.

huangxichen1:
This PR is to fix the bug when the cardinality of two sketches are both 0, the estimated intersection will be `nan`.


jiayu-google (approved):
Thank you Xichen. Good catch.

jiayu-google (commented):
Chatted with Xichen offline, request some modification.

kungfucraig (approved):

PR #75, opened 2020-09-24
Remove pickle.

huangxichen1:
Reasons to remove pickle:
1. It is not used by the output.
2. Pickle can only serialize the top level class and functions. And in this code base, we are using `factory` methods a lot, and it is easy to get errors. For example, the stratified sketch VoC and stratified sketch ADBF use sketch_factory, so they are not compatible with pickle. The error message looks like:
```
AttributeError: Can't pickle local object 'VectorOfCounts.get_sketch_factory.<locals>.f'
```
where `<locals>` is a class referred by the `get_sketch_factory` method.

### DIFF
--- a/src/evaluations/evaluator.py
+++ b/src/evaluations/evaluator.py
@@ -16,7 +16,6 @@
 import copy
 import itertools
 import os
-import pickle
 import shutil
 import time
 
@@ -29,11 +28,7 @@ from wfa_cardinality_estimation_evaluation_framework.evaluations.configs import 
 from wfa_cardinality_estimation_evaluation_framework.evaluations.configs import SketchEstimatorConfig
 from wfa_cardinality_estimation_evaluation_framework.simulations.simulator import Simulator
 
-# Pickle filenames for saving the configurations.
-SCENARIO_SEED_FILE = 'set_generator_seed.p'
-SCENARIO_CONFIG_FILE = 'scenario_config.p'
-ESTIMATOR_CONFIG_FILE = 'sketch_estimator_config.p'
-RUN_DIRS_FILE = 'run.p'
+# Filenames.
 EVALUATION_RUN_TIME_FILE = 'evaluation_run_time'
 
 # Keys of the dictionary to remember where the configs and results are saved.
@@ -266,9 +261,6 @@ class Evaluator(object):
         out_dir=out_dir,
         overwrite=overwrite)
 
-    with open(os.path.join(out_dir, run_name, RUN_DIRS_FILE), 'wb') as f:
-      pickle.dump(self.description_to_file_dir, f)
-
     # This is to make sure that different estimators use the same simulation
     # data under the same scenario.
     scenario_random_states = {}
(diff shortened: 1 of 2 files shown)


jgoodknight:
This is interesting... why wasn't it a problem before?  I also believe this is why Preston used the pathos multiprocessing library instead of Python's built-in which has to pickle everything it calculates.  There's a library in that package called dill that will do the same thing as pickle but works on lambdas you could use instead of getting rid of the pickling if we need it

huangxichen1:
> This is interesting... why wasn't it a problem before? I also believe this is why Preston used the pathos multiprocessing library instead of Python's built-in which has to pickle everything it calculates. There's a library in that package called dill that will do the same thing as pickle but works on lambdas you could use instead of getting rid of the pickling if we need it

The reason that we didn't see this problem before is that we try our best to avoid lower level class calls. For example, I made some change to the Liquid Legion code to make the LiquidLegion pickle-able.

Yet, it becomes more difficult when we do the frequency evaluation. The frequency estimators may depend on the cardinality estimators and use get_sketch_factory method of the cardinality estimators, which will cause the issue that the class is called on a lower level. For example, the [stratified sketch](https://github.com/world-federation-of-advertisers/cardinality_estimation_evaluation_framework/blob/master/src/estimators/stratified_sketch.py#L77) uses `cardinality_sketch_factory`. Let's say we use VoC, and the pickle will raise `AttributeError: Can't pickle local object 'VectorOfCounts.get_sketch_factory.<locals>.f'`.

Let me try the `dill`, and report back if it is possible.

huangxichen1:
> > This is interesting... why wasn't it a problem before? I also believe this is why Preston used the pathos multiprocessing library instead of Python's built-in which has to pickle everything it calculates. There's a library in that package called dill that will do the same thing as pickle but works on lambdas you could use instead of getting rid of the pickling if we need it
> 
> The reason that we didn't see this problem before is that we try our best to avoid lower level class calls. For example, I made some change to the Liquid Legion code to make the LiquidLegion pickle-able.
> 
> Yet, it becomes more difficult when we do the frequency evaluation. The frequency estimators may depend on the cardinality estimators and use get_sketch_factory method of the cardinality estimators, which will cause the issue that the class is called on a lower level. For example, the [stratified sketch](https://github.com/world-federation-of-advertisers/cardinality_estimation_evaluation_framework/blob/master/src/estimators/stratified_sketch.py#L77) uses `cardinality_sketch_factory`. Let's say we use VoC, and the pickle will raise `AttributeError: Can't pickle local object 'VectorOfCounts.get_sketch_factory.<locals>.f'`.
> 
> Let me try the `dill`, and report back if it is possible.

Joey: I just tested that the `dill` library can pickle the configs, which is promising.

So now there are two options:
1. Change `pickle` to `dill`, in case we need the serialized objects in the future.
2. Remove `pickle`, as it is not used by any other modules in the framework.

@all, please let me know which one you prefer.

kungfucraig (approved):

jgoodknight (approved):

my2457 (approved):

PR #62, opened 2020-09-04
[Frequency analyzer part 4] Save frequency distribution plot

huangxichen1:

PR #61, opened 2020-09-04
[Frequency analyzer part 3] Convert the raw_df to long format

huangxichen1:
This is to provide data for generating barplot for comparing frequency distributions.

PR #76, opened 2020-09-24
[Fix] Create per frequency plot.

huangxichen1:
The previous PR #61 and PR #62 made frequency plots, which is cumulative instead of per-frequency plot. This PR fixes this bug, and creates per frequency distribution plots.

### DIFF
--- a/src/evaluations/analyzer.py
+++ b/src/evaluations/analyzer.py
@@ -381,7 +381,7 @@ class FrequencyEstimatorEvaluationAnalyzer(EstimatorEvaluationAnalyzer):
        frequency_level  cardinality
                      1            2
                      2            1
-                     3            4
+                     3            7
     where the cardinality of frequency_level == 1 is the number of IDs that have
     frequency equal to 1.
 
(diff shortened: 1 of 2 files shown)


src/evaluations/analyzer.py
```diff
@@ -366,6 +366,64 @@ def convert_raw_df_to_long_format(self):
...
+
+    For example, given a table of cumulative-frequency cardinality,
+       frequency_level  cardinality
+                     1           10
+                     2            8
+                     3            7
+    where the cardinality of frequency_level == 1 means that the number of IDs
+    that have at least 1 impressions, running this method will return
+       frequency_level  cardinality
+                     1            2
```
jiayu-google, line 382: maybe write out 10 - 8 = 2, or just 10 - 8

src/evaluations/analyzer.py
```diff
@@ -366,6 +366,64 @@ def convert_raw_df_to_long_format(self):
...
+                     2            1
+                     3            4
```
jiayu-google, line 384: 4 -> 7

jiayu-google (approved):
Thanks Xichen!

my2457 (approved):

PR #78, opened 2020-09-25
Fix a tiny bug of UniqueKeyFunction()

jiayu-google:


huangxichen1 (approved):
We're submitting a quick bug fix. Feel free to review after submission if you'd like.

PR #80, opened 2020-09-26
Make the universe_size variable.

huangxichen1:


jiayu-google (approved):
Nice to have the feature of varying the universe! Since finally people are interested in the performance in the real universe, and we need to figure out how the universe size affects the performance.

src/evaluations/data/evaluation_configs.py
```diff
@@ -80,7 +80,7 @@
 
 
 # Document the evaluation configurations.
-def _smoke_test(num_runs=NUM_RUNS_VALUE):
+def _smoke_test(num_runs=NUM_RUNS_VALUE, universe_size=200000):
```
shengma-1, line 83: 2e5 is in several places. Does it make sense to create a global variable: DEFAULT_UNIVERSE_SIZE?

src/evaluations/data/evaluation_configs.py
```diff
@@ -153,24 +158,25 @@ def _frequency_smoke_test(num_runs=NUM_RUNS_VALUE):
...
               name='homogeneous',
               set_generator_factory=(
-                frequency_set_generator.HomogeneousMultiSetGenerator.
-                get_generator_factory_with_num_and_size(
-                    universe_size=int(2e5), num_sets=10, set_size=int(2e4),
-                    freq_rates=[1]*10, freq_cap=3))),
+                  frequency_set_generator.HomogeneousMultiSetGenerator
+                  .get_generator_factory_with_num_and_size(
+                      universe_size=universe_size, num_sets=10,
+                      set_size=int(2e4), freq_rates=[1]*10, freq_cap=3))),
```
jgoodknight, line 164: doing this list multiplication trick has bitten me in the butt before but you probably won't get in trouble since you won't be modifying the list?  if I remember correctly the issue has to do with getting 10 copies of the same object instead of 10 new objects, so also with integers it's less of a problem.  That being said, in general ever since then I  just do a list comprehension instead [1 for _ in range(10)] so there's no chance of getting bitten.  This is just an FYI, no need to fix it here unless you want to

jgoodknight (commented):
I woul

src/evaluations/data/evaluation_configs.py
huangxichen1: Done.

src/evaluations/data/evaluation_configs.py
```diff
@@ -137,10 +141,11 @@ def _smoke_test(num_runs=NUM_RUNS_VALUE):
       )
 
 
-def _frequency_smoke_test(num_runs=NUM_RUNS_VALUE):
+def _frequency_smoke_test(num_runs=NUM_RUNS_VALUE, universe_size=int(2e5)):
```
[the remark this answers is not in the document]
huangxichen1: Done. Great to know this magic. Thanks Joey!

src/evaluations/data/evaluation_configs.py
huangxichen1: Acknowledged. I will not change in this PR. I will follow list comprehension in the future PRs. Thanks!

src/evaluations/data/evaluation_configs.py
```diff
@@ -89,10 +91,13 @@ def _smoke_test(num_runs=NUM_RUNS_VALUE):
 
   Args:
     num_runs: the number of runs per scenario.
+    universe_size: the size of universe.
 
   Returns:
     An EvaluationConfig.
   """
+  set_size = int(universe_size * 0.2)
```
jiayu-google, line 99: Maybe set 0.2 and 0.1 as global variables?

src/evaluations/data/evaluation_configs.py
```diff
@@ -629,11 +641,10 @@ def _complete_test_with_selected_parameters(
       scenario_config_list=scenario_config_list)
 
 
-def _frequency_end_to_end_test(num_runs=NUM_RUNS_VALUE):
+def _frequency_end_to_end_test(universe_size=10000, num_runs=NUM_RUNS_VALUE):
   """EvaluationConfig of end-to-end test of frequency evaluation code."""
   num_sets = 3
-  universe_size = 10000
-  set_size = 5000
+  set_size = int(universe_size * 0.5)
```
jiayu-google, line 647: Again, may call out 0.5 somewhere.

jiayu-google (approved):
Thanks Xichen. I'm not sure whether we should call out some parameters as global variables.
To me, the reach proportions (for which you used 0.1, 0.2 and 0.5 here) are worth calling out.

src/evaluations/data/evaluation_configs.py
```diff
@@ -44,6 +44,7 @@
     MAX_FREQUENCY)
 
 NUM_RUNS_VALUE = 100
+SMOKE_TEST_UNIVERSE_SIZE = 200_000
 UNIVERSE_SIZE_VALUE = 1000000
```
jgoodknight, line 48: can you change this file to consistently use underscores for all large numbers?  Not that big a deal but when I went to review it stuck out

tests/interoperability_test.py
```diff
@@ -417,6 +417,7 @@ def test_run_evaluation_for_cardinality_estimator_workflow_compatible(self):
         sketch_estimator_configs=sketch_estimator_configs,
         evaluation_run_name='interoperability_test_for_evaluator_cardinality',
         num_runs=1,
+        universe_size=int(1000),
```
jgoodknight, line 420: int definitely not needed here

tests/interoperability_test.py
```diff
@@ -451,6 +452,7 @@ def test_run_evaluation_for_frequency_estimator_workflow_compatible(self):
         sketch_estimator_configs=sketch_estimator_configs,
         evaluation_run_name='interoperability_test_for_evaluator_frequency',
         num_runs=1,
+        universe_size=int(1000),
```
jgoodknight, line 455: int not needed here either

jgoodknight (approved):

shengma-1 (approved):

tests/interoperability_test.py
huangxichen1: Done.

tests/interoperability_test.py
huangxichen1: Done.

src/evaluations/data/evaluation_configs.py
huangxichen1: Done.

my2457 (approved):

PR #83, opened 2020-09-30
Bug fix for report generator

huangxichen1:
Fix for plot link missing due to wrong sketch estimator name.

The bug is from the difference of the sketch_estimator name of the cardinality estimator and the frequency estimator. The previous code will attach -nan to the sketch_estimator name, and that name will not be found in the directory tree. So this PR  does the following:
1. Fix this bug by checking the analysis type.
2. Add testing for plot file link in the HTML file.


AaronRanAn (approved):
LGTM!

jgoodknight (approved):

PR #85, opened 2020-10-01
Minor change to analyzer.py

huangxichen1:
1. Close fig after saving plot. If without this fix, will see the following message when running the complete evaluation:
```
RuntimeWarning: More than 20 figures have been opened. Figures created through the pyplot interface (`matplotlib.pyplot.figure`) are retained until explicitly closed and may consume too much memory. (To control this warning, see the rcParam `figure.max_open_warning`)
```
2. Fix a typo of a global variable value.

### DIFF
--- a/src/evaluations/analyzer.py
+++ b/src/evaluations/analyzer.py
@@ -16,6 +16,7 @@
 import os
 import shutil
 
+import matplotlib.pyplot as plt
 import numpy as np
 import pandas as pd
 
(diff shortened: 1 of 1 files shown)


my2457 (approved):

src/evaluations/analyzer.py
```diff
@@ -456,6 +458,7 @@ def plot_one_estimator_under_one_scenario(df):
           w=self.plot_params[BARPLOT_SIZE_WIDTH_INCH],
           h=self.plot_params[BARPLOT_SIZE_HEIGHT_INCH])
       fig.savefig(plot_file)
+      plt.close(fig)
```
kungfucraig, line 461: I'd recommend using a context manager for this.

kungfucraig (approved):

src/evaluations/analyzer.py
jgoodknight: good idea

jgoodknight (approved):

src/evaluations/analyzer.py
huangxichen1: Per our offline discussion, I will leave this as is.

PR #86, opened 2020-10-05
Fix a tiny bug in bloom_filter_sketch_operators

jiayu-google:


my2457 (approved):
Thanks for the fix

shengma-1 (approved):

huangxichen1 (approved):
Thanks, Jiayu!

PR #88, opened 2020-10-08
Fix clipping method of VoC.

jiayu-google:
Fix a bug in clip_empty_vector_of_count, and add tests.

Note: Sorry that I did not test this part seriously when first writing the clipping method. Other parts of the clipping method had been tested thoroughly and should have no bugs.


huangxichen1 (approved):
Thanks Jiayu!

If easy to do, would you mind adding tests for clipping the intersection?

jiayu-google:
Thanks Xichen! Added more tests on intersection.

Note: clipping intersection was implemented using the method evaluate_closeness_to_a_value. The design was not consistent with clip_empty_vector_of_count. If needed, I can try to refactor the codes.

my2457 (approved):

PR #91, opened 2020-10-09
Update geometric bloom filter for better accurancy

my2457:
Added GEO_LENGTH_PROB_PRODUCT to allow config to automatically adjust probability that can give a fairly accurate result


huangxichen1 (approved):
Thank you, Miao!

PR #96, opened 2020-10-22
update HLL++ to use sparse mode

jgoodknight:
HLL++ doesn't do well for small cardinalities so we are updating our implementation to use a version of sparse mode from the HLL++ paper, that doesn't care so much about memory space but is much simpler to implement.

Note this is not tested.

### DIFF
--- a/src/estimators/tests/hyper_log_log_test.py
+++ b/src/estimators/tests/hyper_log_log_test.py
@@ -157,6 +157,65 @@ class HyperLogLogPlusPlusTest(absltest.TestCase):
   def test_insert_huge(self):
     self.insertion_test_helper(1_000_000)
 
+  def test_merge_sparse_with_sparse_to_sparse(self):
+    hll1 = HyperLogLogPlusPlus(length=16, random_seed=234)
+    hll1.add(1)
+    hll2 = HyperLogLogPlusPlus(length=16, random_seed=234)
+    hll2.add(1)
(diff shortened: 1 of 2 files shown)


kungfucraig (changes requested):
Please add tests, especially for the merge.

huangxichen1:
> Please add tests, especially for the merge.

Added tests.

> Please add tests, especially for the merge.

I added the tests. PTAL. Thanks!

src/estimators/tests/hyper_log_log_test.py
```diff
@@ -155,6 +157,65 @@ def test_insert_large(self):
...
+
+    merged_hll = hll1.merge(hll2)
+    self.assertTrue(merged_hll.sparse_mode,
+                    'Merged sketch should be in sparse mode.')
+
+    hll1.add(1000)
+    merged_hll = hll1.merge(hll2)
+    self.assertFalse(merged_hll.sparse_mode,
+                     'Merged sketch should not be in sparse mode.')
+
```
kungfucraig, line 187: Can you test the cardinality of the merged hll is as expected?

kungfucraig (approved):

src/estimators/tests/hyper_log_log_test.py
huangxichen1: I added tests for the estimated cardinality.

PR #101, opened 2020-10-30
Add more parameters for Exp ADBF.

huangxichen1:
This PR adds more parameters to the Exp-ADBF config, so that it can match with the Geometric-ADBF parameter settings.

### DIFF
--- a/src/evaluations/data/evaluation_configs.py
+++ b/src/evaluations/data/evaluation_configs.py
@@ -86,7 +86,7 @@ ESTIMATE_EPSILON_VALUES = (math.log(3), None)
 # We use the np.array with dtype so as to make sure that the lengths are all
 # integers.
 ADBF_LENGTH_LIST = np.array([100_000, 250_000], dtype=np.int64)
-EXP_ADBF_DECAY_RATE = 10
+EXP_ADBF_DECAY_RATE_LIST = np.array([2, 10], dtype=np.int64)
 STRATIFIED_EXP_ADBF_EPSILON_SPLIT = 0.5
 SKETCH_OPERATOR_EXPECTATION = 'expectation'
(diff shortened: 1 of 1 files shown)


huangxichen1:
@my2457 @AaronRanAn 
Now I am setting the baseline branch as `geo_bf_config_update_2` so as to show the additional change from PR #95. I do not intend to merge this PR with PR #95. Instead, I will rebase it against the master branch after you merge PR #95 to the master. So please go ahead and merge PR #95 to the master once you are ready.

src/evaluations/data/evaluation_configs.py
```diff
@@ -86,7 +86,7 @@
 # We use the np.array with dtype so as to make sure that the lengths are all
 # integers.
 ADBF_LENGTH_LIST = np.array([100_000, 250_000], dtype=np.int64)
-EXP_ADBF_DECAY_RATE = 10
+EXP_ADBF_DECAY_RATE_LIST = np.array([2, 10], dtype=np.int64)
```
kungfucraig, line 89: Do we care about decay_rate < 10?

kungfucraig (changes requested):

src/evaluations/data/evaluation_configs.py
huangxichen1: Here we consider decay rate equal to 2 or 10.

When decay rate = 2, the exp-ADBF and the geo-ADBF config in this framework would be equivalent.
Decay rate = 10 is proposed to be used in a wide range of scenarios.

If other decay rates are needed, we can add them later.

my2457 (approved):
lgtm

TNATALI:
No longer valid

PR #102, opened 2020-11-03
[Part 1 Liquid Legion Global DP tests]Add disjoint set generator.

huangxichen1:
This chain of PRs is to test the limit of liquid legions under different epsilon * audience sizes combinations.

This PR is to add a disjoint set generator which is a deterministic so that the simulation could be much faster.

### DIFF
--- a/src/simulations/set_generator.py
+++ b/src/simulations/set_generator.py
@@ -523,8 +523,8 @@ class DisjointSetGenerator(SetGeneratorBase):
 
   def __iter__(self):
     for set_size in self.set_sizes:
-      set_ids = np.arange(self.start_id, self.start_id + set_size, dtype=int)
-      self.union_ids = self.union_ids.union(set_ids)
+      set_ids = range(self.start_id, self.start_id + set_size)
+      self.union_ids = range(self.start_id+set_size)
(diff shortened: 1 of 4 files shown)


src/simulations/set_generator.py
```diff
@@ -486,3 +486,45 @@ def __iter__(self):
...
+        conform to the set generator constructor arguments.
+    """
+    self.union_ids = set()
+    _ = random_state
+    self.set_sizes = set_sizes
+    self.start_id = 0
+
+  def __iter__(self):
+    for set_size in self.set_sizes:
+      set_ids = np.arange(self.start_id, self.start_id + set_size, dtype=int)
```
kungfucraig, line 526: Why not use native "range"?

Also, you should write something like this:

out = range(self.start_id, self.start_id+set_size)
self.union_ids = range(0, self.start_id+set_size)
self.start_id += set_size
return out

This will conserve memory.

kungfucraig (changes requested):

src/simulations/set_generator.py
huangxichen1: Done.

kungfucraig (approved):

PR #105, opened 2020-11-04
Initial commit for GaussianEstimateNoiser

pasin30055:
Implement GaussianMechanism and GeometricEstimateNoiser, which add Gaussian noise to a function/estimator to ensure differential privacy.

Note: currently the noise is calculated using a bound from Dwork-Roth book which is only valid for epsilon no more than one. A TODO is added to use a more accurate computation of sigma, which is also valid for all ranges of epsilon and delta.

### DIFF
--- a/src/common/noisers.py
+++ b/src/common/noisers.py
@@ -44,10 +44,7 @@ class LaplaceMechanism:
     self._func = f
     self._delta_f = delta_f
     self._epsilon = epsilon
-    if random_state:
-      self._random_state = random_state
-    else:
-      self._random_state = np.random.RandomState()
+    self._random_state = random_state or np.random.RandomState()
 
   def __call__(self, x):
     z = self._func(x)
@@ -103,10 +100,7 @@ class GeometricMechanism:
     self._func = f
     self._delta_f = delta_f
     self._epsilon = epsilon
-    if random_state:
-      self._random_state = random_state
-    else:
-      self._random_state = np.random.RandomState()
+    self._random_state = random_state or np.random.RandomState()
 
   def __call__(self, x):
     z = self._func(x)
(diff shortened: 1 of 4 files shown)


src/common/noisers.py
```diff
@@ -116,6 +116,52 @@ def __call__(self, x):
...
+      random_state:  Optional instance of numpy.random.RandomState that is
+        used to seed the random number generator.
+    """
+    self._func = f
+    self._delta_f = delta_f
+    # TODO(pasin30055): Use tight computation of sigma.
+    # Set the standard deviation sigma, following Appendix A of Dwork and Roth.
+    self._sigma = (math.sqrt(2 * math.log(1.25 / delta)) * delta_f *
+      math.sqrt(num_queries) / epsilon)
+    if random_state:
```
kungfucraig, line 155: self.random_state = random_state if random_state is not None else RandomState()

kungfucraig (commented):
I think you mean GaussianEstimateNoiser, right?

pasin30055:
> I think you mean GaussianEstimateNoiser, right?

src/common/noisers.py
pasin30055: Thanks for the suggestion. I changed this to a more compact:
  self._random_state = random_state or np.random.RandomState()
and also made similar changes in LaplaceMechanism, GeometricMechanism.

pasin30055:
> I think you mean GaussianEstimateNoiser, right?

I'm not sure whether you're referring to the title of the pull request, but I've changed it to GaussianEstimateNoiser. If you meant something else, please let me know.

jiayu-google (approved):
LGTM, thanks Pasin!

pasin30055:
@jiayu-google Would you mind merging this request? (I don't have the write access...)

my2457 (approved):

jiayu-google:
@pasin30055  Just merged.

PR #114, opened 2020-11-17
Use tight Gaussian noise

pasin30055:
This changes make the amount of Gaussian noise we add the smallest possible that still provides the given privacy level.
(The computation of the noise is done via the [dp-accounting](https://pypi.org/project/dp-accounting/) package.)


jiayu-google (approved):
Thanks Pasin!

PR #120, opened 2021-05-05
Refactor StandardizedHistogramEstimator

jiayu-google:
Update StandardizedHistogramEstimator to (i) include the delta parameter (ii) support privacy budget tracking (iii) add more tests.

Will add more tests once Pasin updates Gaussian and DiscreteGaussian noisers. 

@matthewclegg I'm trying to support privacy budget tracking here, but not sure if my code design is reasonable.

@pasin30055 FYI of what we're doing. No need to carefully review it at this moment. An open question here is: how to composite the privacy budget for both reach and frequency estimation? You have result for discrete Laplacian noises in our paper, but how about Gaussian noises? We can discuss more in Tuesday's meeting.

### DIFF
--- a/src/estimators/same_key_aggregator.py
+++ b/src/estimators/same_key_aggregator.py
@@ -15,7 +15,9 @@
 
 import copy
 import functools
+import inspect
 import numpy as np
+import warnings
 
 from wfa_cardinality_estimation_evaluation_framework.estimators import any_sketch
 from wfa_cardinality_estimation_evaluation_framework.estimators.any_sketch import UniqueKeyFunction
@@ -24,6 +26,8 @@ from wfa_cardinality_estimation_evaluation_framework.estimators.base import Sket
 from wfa_cardinality_estimation_evaluation_framework.estimators.bloom_filter_sketch_operators import SketchOperator
 from wfa_cardinality_estimation_evaluation_framework.estimators.bloom_filters import ExponentialBloomFilter
 from wfa_cardinality_estimation_evaluation_framework.estimators.bloom_filters import FirstMomentEstimator
+from wfa_cardinality_estimation_evaluation_framework.estimators.estimator_noisers import DiscreteGaussianEstimateNoiser
+from wfa_cardinality_estimation_evaluation_framework.estimators.estimator_noisers import GaussianEstimateNoiser
 from wfa_cardinality_estimation_evaluation_framework.estimators.estimator_noisers import GeometricEstimateNoiser
 
 
@@ -94,23 +98,38 @@ class ExponentialSameKeyAggregator(SketchBase):
 
 class StandardizedHistogramEstimator(EstimatorBase):
   """Frequency estimator from ExponentialSameKeyAggregator.
+
+    Algorithm Description
+    (following the argument names in __init__ and __call__):
+    Given any sketch_list which includes the ExponentialSameKeyAggregator from
+    a set of publishers, we estimate the frequency histogram among all the
+    publishers in the following steps.
+    Step 1. Estimate the one_plus_reach, which is 1+ union reach of all pubs.
+    Step 2. Estimate the active_key_histogram, which is the histogram of
+      the frequency for each active key, i.e., each id that is the unique
+      reached id in its bucket.
+    Step 3. Noise one_plus_reach by
+      noiser_class(reach_epsilon, **reach_noiser_kwargs).
+      And noise active_key_histogram by
+      noiser_class(frequency_epsilon, **frequency_noiser_kwargs).
+    Step 4. Return noised_one_plus_reach * noised_active_key_histogram /
+      sum(noised_active_key_histogram).
   """
 
   def __init__(self,
                max_freq=10,
-               noiser_class=GeometricEstimateNoiser,
-               epsilon=np.log(3),
-               epsilon_split=0.5):
+               reach_noiser_class=GeometricEstimateNoiser,
+               frequency_noiser_class=GeometricEstimateNoiser,
+               reach_epsilon=1.,
+               frequency_epsilon=1.,
+               reach_delta=0.,
+               frequency_delta=0.,
+               reach_noiser_kwargs={},
+               frequency_noiser_kwargs={}):
     """Initiate a StandardizedHistogramEstimator.
 
-    Algorithm description:
-    Given any ExponentialSameKeyAggregator ska,
-    Step 1. Estimate the 1+ reach, from ska.exponential_bloom_filter.
-    Step 2. Estimate the histogram of frequency histogram among (only)
-      effective keys.
-    Step 3. Use the estimated 1+ reach to standardize the frequency histogram
-      among effective keys, and thus obtain an estimate of frequnecy histogram
-      among all IDs.
+    To understand the usage of the arguments here, please read the
+    "Algorithm Description" in the docstring of the whole class.
 
     Args:
       max_freq: the maximum targeting frequency level. For example, if it is set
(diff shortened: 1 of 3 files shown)


src/estimators/same_key_aggregator.py
```diff
@@ -94,54 +98,107 @@ def assert_compatible(self, other):
...
+      a tuple of (<obtained noiser>, <actual epsilon the noiser guarantees>,
+      <actual delta the noiser guarantees>)
+    """
+    if noiser_class is None:
+      return (None, np.Inf, 0)
+    if delta in inspect.getfullargspec(noiser_class)[0]:
+      # If delta is an argument of noiser_class
+      return (noiser_class(epsilon=epsilon, delta=delta, **kwargs),
+              epsilon, delta)
+    # Otherwise, delta is not an argument of noiser_class
```
matthewclegg, line 191: Maybe check if delta is zero before issuing this warning?

src/estimators/same_key_aggregator.py
```diff
@@ -94,54 +98,107 @@ def assert_compatible(self, other):
...
+    """A util for defining reach_noiser and frequency_noiser.
+
+    Args:
+      noiser_class: given name of noiser_class.
+      epsilon: given epsilon for the noiser.
+      delta: given delta for the noiser.
+      kwargs: other args to be specified in the noiser_class.
+
+    Returns:
+      a tuple of (<obtained noiser>, <actual epsilon the noiser guarantees>,
```
matthewclegg, line 182: What is the relationship between the epsilon that is given as an input value versus the epsilon guaranteed by the noiser?  For example, is it the case that the guaranteed epsilon <= input epsilon?

src/estimators/same_key_aggregator.py
```diff
@@ -94,54 +98,107 @@ def assert_compatible(self, other):
...
+      return (None, np.Inf, 0)
+    if delta in inspect.getfullargspec(noiser_class)[0]:
+      # If delta is an argument of noiser_class
+      return (noiser_class(epsilon=epsilon, delta=delta, **kwargs),
+              epsilon, delta)
+    # Otherwise, delta is not an argument of noiser_class
+    warnings.warn('The given delta is ignored, since it is not in the noiser.')
+    return (noiser_class(epsilon=epsilon, **kwargs), epsilon, 0)
+
+  def calculate_total_privacy_budget(self):
```
matthewclegg, line 195: Rather than calculating the total privacy consumption and returning it, I think it would be better to return a list of (epsilon, delta) pairs representing the noise amounts that were added to the data for the different noising events.  These different noising events could then be individually recorded in the PrivacyTracker object in the planning eval framework.  This would facilitate using the advanced composition theorem to calculate the total privacy usage over the run of a modeling strategy (or multiple runs of modeling strategies).  Also, it would centralize the computation of privacy consumption in a single location in the code.  So, if we later decide to add support for PLD, only one place in the code will have to be changed.  What do you think?

src/estimators/same_key_aggregator.py
```diff
@@ -94,54 +98,107 @@ def assert_compatible(self, other):
...
+    return (self.reach_epsilon + self.frequency_epsilon * self.max_freq,
```
matthewclegg, line 196: This code seems to suggest that the parallel composition theorem cannot be applied to the different buckets in the frequency histogram.  Is this true?  To me, it seems that if one user is added or removed, that user can only affect a single frequency bucket.  Am I wrong about this?

src/estimators/same_key_aggregator.py
```diff
@@ -94,54 +98,107 @@ def assert_compatible(self, other):
 
 class StandardizedHistogramEstimator(EstimatorBase):
   """Frequency estimator from ExponentialSameKeyAggregator.
+
```
matthewclegg, line 101: Very clear documentation!  Thank you!!

matthewclegg (changes requested):

src/estimators/same_key_aggregator.py
jiayu-google: Quick answer to this comment (will answer other comments later):
The simple parallel composition assumes delta = 0. It does not work for Gaussian noiser which must have delta > 0. @pasin30055 Is there an efficient parallel composition for delta > 0?

Perhaps, we can use Discrete Laplace for active_key_histogram (to have parallel composition) and Discrete Gaussian for one_plus_reach (since Gaussian is efficient for composition of large # of queries).

jiayu-google: Sorry for the complexity here. It should be easy to just derive the correct parallel composition for the case with delta. Let me do it over the weekend and hopefully through the math we'll see how it is different from the simple parallel composition.

src/estimators/same_key_aggregator.py
pasin30055, line 182: What is an example when this is smaller than the input epsilon?
(I can imagine the delta case where delta = 0 for Laplace or something like that but not surer about epsilon.)

src/estimators/same_key_aggregator.py
pasin30055: +1 regarding noising events as those might allow us to do tighter compositions

src/estimators/same_key_aggregator.py
pasin30055: A couple of comments:
- I don't see why parallel composition wouldn't work for GaussianMechanism. Can you please clarify?
- I'd actually support letting the histogram noiser itself deals with its own epsilon, delta. I.e. let's pass in the epsilon, delta and let the noiser computes the noise to be added to each bucket to achieve the input epsilon, delta. This way we can remove the "* self.max_freq" here and possibly in the future support other more complicated noisers. Wdyt?

pasin30055 (commented):
Thanks Jiayu. I left a few additional comments (mostly on issues already raised by Matthew).

jiayu-google (commented):
Thank you for your careful review and great suggestions, Matthew and Pasin.
Revised, and could you please review again.

pasin30055 (changes requested):
Thanks Jiayu. The changes look good; a few additional comments below.

src/estimators/same_key_aggregator.py
```diff
@@ -94,54 +98,126 @@ def assert_compatible(self, other):
...
+      return noiser_class(epsilon=epsilon, **kwargs), True
+    return noiser_class(epsilon=epsilon, **kwargs), False
+
+  def ouput_privacy_parameters(self):
+    """Returns a list of (eps, delta) for different noising events.
+
+    Different entries of this list will be later composed, say, using advanced
+    composition theorem.
+    """
+    return [(self.reach_epsilon, self.reach_delta),
```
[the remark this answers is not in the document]
matthewclegg: That would be ideal.

matthewclegg (changes requested):

src/estimators/same_key_aggregator.py
```diff
@@ -94,54 +98,126 @@ def assert_compatible(self, other):
...
+                          else reach_epsilon)
+    self.frequency_epsilon = (np.Inf if frequency_noiser_class is None
+                              else frequency_epsilon)
+    self.reach_delta = reach_delta
+    self.frequency_delta = frequency_delta
+    self.one_plus_reach_noiser, ignore_delta = (
+        StandardizedHistogramEstimator.define_noiser(
+            reach_noiser_class,
+            reach_epsilon, reach_delta, reach_noiser_kwargs))
+    if ignore_delta:
```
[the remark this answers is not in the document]
jiayu-google: A complexity here is that for frequency noiser, it will return per_bucket_delta and we should multiply it by 2 to save the whole histogram delta. While for reach we directly save the returned delta.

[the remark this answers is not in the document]
matthewclegg: ok

matthewclegg (approved):

PR #123, opened 2021-06-14
Add HeaviestOverlapGenerator that will be used in planning_evaluation

jiayu-google:


src/simulations/set_generator.py
```diff
@@ -79,6 +79,47 @@ def __iter__(self):
     return self
```
matthewclegg, line 79: Hi Jiayu,  

Greetings from Oulu, Finland!

I'm not sure that this PR is needed, and I was wondering if you could explain further.  If you look at the `lhs_data_design_example.py` that is in PR #30 ([Enables synthetic_data_design_generator to read designs directly from files](https://github.com/world-federation-of-advertisers/planning-evaluation-framework/pull/30), you will see the following code fragment:

```
OVERLAP_GENERATORS = [
    GeneratorParameters("FullOverlap", DataSet, {}),
    GeneratorParameters(
        "Independent",
        IndependentOverlapDataSet,
        {"universe_size": 50_000, "random_generator": 1},
    ),
...
```

The `FullOverlap` overlap generator is intended to be the same as the `HeaviestOverlapGenerator`, but it does not require creating any additional classes.  What do you think?  Is the `HeaviestOverlapGenerator` needed for another reason?  Or did I get something wrong with the implementation of `FullOverlap`?

Matt

matthewclegg (changes requested):

src/simulations/set_generator.py
jiayu-google: Greetings from New Have CT:) I'm having a shorter trip than you --- first to New Hampshire and then Connecticut.

The `FullyOverlapSetGenerator` class assumes equal size for each set, so it can never take a `set_size_list`. Changing this class will affect a few other places like evaluation_configs.py.

I guess you would prefer deleting `HeaviestOverlapGenerator`, changing the `FullyOverlapSetGenerator` class and also the codes using it. LMK if you have other suggestions!

jiayu-google: Matthew, now I understand your point! Sorry for being slow... Will delete this PR then.

PR #125, opened 2021-08-04
Fixes deprecation warning in common/random.py and removes warning in …

matthewclegg:
…estimators/same_key_aggregator.py


TNATALI:
No longer valid

PR #132, opened 2022-04-15
Open Sourcing Differential Privacy Training Docs

jgoodknight:
I created these docs in the course of my work on differential privacy at google, and they have proven to be useful tools to educate people on differential privacy.  In the interest of serving the mission of spreading differential privacy education, and the WFA's mission, we thought we would open-source these docs so they can be useful beyond just Google.


jgoodknight:
wrong repo!

